PR #3, opened 2016-04-08
Fix various small issues

desilinguist:
- [RSMR-150] Add more useful system information
- [RSMR-162] Improve type handling in `preprocess.remove_outliers()`
- [RSMR-113] Allow `candidate_column` and `id_column` to have the same values.
- Some minor clean up here and there
- Tell CircleCI to cache the conda packages so that they don't have to be downloaded every time.


rsmtool/input.py
```diff
@@ -860,15 +860,24 @@ def load_and_filter_data(csv_file,
...
+    # read in the CSV file
     df = pd.read_csv(csv_file, converters=converter_dict)
 
+    # it is possible for the `id_column` and `candidate_column` to be
+    # set to the same column name in the CSV file, e.g., if there is
+    # only one response per candidate. If this happens, we neeed to
+    # create a duplicate column for candidates for the downstream
+    # processing to work as usual.
+    if id_column == candidate_column:
+        df['candidate'] = df['ID'].copy()
```
aloukina: This should <<df[id_column]>> (the test pass because this is what you used as ID column)

I think we should have this bit (877-879) after checking that all specified names for all columns present in the data (that is between <<if missing columns:>> and <<df=rename...>>, else the tool will raise an obscure ValueError if the user gave wrong column name in both cases.

rsmtool/notebooks/header.ipynb
```diff
@@ -33,6 +33,7 @@
     "import os\n",
     "import re\n",
     "import pickle\n",
+    "import platform\n",
```
aloukina: Does this need to go into conda requirements?

rsmtool/notebooks/sysinfo.ipynb
```diff
@@ -8,6 +8,31 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "system_name = platform.system()\n",
+    "\n",
+    "# People might not know what 'Darwin' is, so we should replace that with 'Mac OS X'\n",
```
aloukina: Nice!

tests/data/experiments/lr-candidate-same-as-id/lr_candidate_same_as_id.json
```diff
@@ -0,0 +1,15 @@
+{
+    "test_label_column": "score",
+    "train_file": "../../files/train.csv",
+    "description": "Using all features with an LinearRegression model.",
```
aloukina: an -> a

aloukina:
The platform works fine on both Ubuntu and Windows :)

rsmtool/input.py
desilinguist: D'oh! Will fix.

Good point. Will move that down.

rsmtool/notebooks/header.ipynb
desilinguist: Nope, it's built in :)

aloukina:
This can be merged once the tests pass

PR #4, opened 2016-04-08
Bugfix/spkitemid and candidate both called candidate

aloukina:
id_column and candidate_column can now both be set to "candidate"


rsmtool/input.py
```diff
@@ -1054,13 +1061,8 @@ def load_and_filter_data(csv_file,
         if exclude_zero_scores:
             df_filtered_human_scores['sc2'] = df_filtered_human_scores['sc2'].replace(0, nan)
 
-    # remove spkitemid from `not_other_columns`
-    # since we want that column in the `other_columns`
-    # data frame
-    not_other_columns.remove('spkitemid')
```
desilinguist: What's going on here?

aloukina: We save all unused columns present in the original data file as "_other_columns" in case people want to use them in special/custom notebooks later. These are the columns listed in 'not_other_columns'. We also add 'spkitemid' for indexing purposes.

I changed the way the df with these columns is set up to make sure that 'spkitemid' is always the first column else whenever the order of the columns change our tests fail because the first column may change and the tests always treat the first column as the index column. Having 'spkitemid' as the first column also makes it easier to work with these files in Excel etc.

desilinguist:
Got it.

PR #7, opened 2016-04-25
Added --version argument

aloukina:


desilinguist:
We should probably set `prog` consistently in all the four scripts. Currently, we use `prog=rsmeval` in `rsmeval` but `prog=rsmtool.py` etc. in the other three tools. This makes the versions strings different.

PR #11, opened 2016-05-24
Comparison notebook modified to include subsection for human human agreement

bndgyawali:


tests/data/experiments/lr-self-compare-with-h2/lr-with-h2/figure/lr_with_h2_betas.svg
```diff
@@ -0,0 +1,1542 @@
...
+  <g id="patch_1">
+   <path d="M 0 216 
+L 576 216 
+L 576 0 
+L 0 0 
+z
+" style="fill:#ffffff;"/>
+  </g>
+  <g id="axes_1">
+   <g id="patch_2">
```
aloukina: Could you remove figures from the test data? Thank you!

rsmtool/notebooks/comparison/score_distributions.ipynb
```diff
@@ -8,6 +8,41 @@
    ]
   },
   {
+   "cell_type": "code",
```
aloukina: I think it's great to have this, but I would argue it should be a separate notebook as I can imagine a case where you may want to exclude this from the report (e.g. nothing changed between the versions).

rsmtool/notebooks/comparison/score_distributions.ipynb
```diff
@@ -8,6 +8,41 @@
...
+   "execution_count": 2,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "old_consistency_file = join(output_dir_old, '{}_consistency.csv'.format(experiment_id_old))\n",
+    "new_consistency_file = join(output_dir_new, '{}_consistency.csv'.format(experiment_id_new))\n",
+    "\n",
+    "if exists(old_consistency_file) and exists(new_consistency_file):\n",
```
aloukina: And another minor consistency issue: we read all other outputs in the header and then pass them onto the other notebooks as a dictionary.

rsmtool/notebooks/comparison/score_distributions.ipynb
bndgyawali: There may not be any change between old and new most of the time. And, as new/old agreement is not comparable, I did not put the difference information here. Though human human agreement is present in rsmtool report, I thought it will be better to have this information in the comparison report too. I may change to have only one row if they are same, but do not think this needs to be excluded. Please let me know.

aloukina: I agree. What I meant is that it would be in the comparison report, but as a separate notebook consistency.ipynb, in the same way as it is handled in rsmtool. This way the user will have control over whether the notebook is in the final report or not. So essentially you would take this code and copy it into a new notebook. It would also require minor changes to report.py (adding the new notebook to the section list).

bndgyawali:
I have addressed the comments by @aloukina. After the build is successful, this branch should be ready for review.

desilinguist:
@bndgyawali this looks good to me! Thanks for helping out! I just did some minor clean-up (fixed some comments and removed a blank cell) in the notebooks that I didn't want to bother you with. 

@aloukina can you look at this and give me an approval and then I will merge.

aloukina:
Looks good to me too. I made a minor edit to change the test name as it was duplicated.
Thank you for following all the conventions!

tests/test_experiment.py
```diff
@@ -1565,7 +1565,7 @@ def test_run_experiment_lr_compare():
     yield check_report, html_report
 
 
-def test_run_experiment_lr_compare():
+def test_run_experiment_lr_compare_with_h2():
```
bndgyawali: thank you @aloukina for catchingthis. I thought I had made the name unique, I somehow missed it.

PR #15, opened 2016-06-03
Feature/rsmr 136 add r squared

aloukina:
- Evaluations on the eval set now include R2 and RMSE
- Additionally, we highlight the main model fit statistics for the training set
- Model fit statistics included into the comparison report
- Updated all test data
- Comparison report no longer breaks for SKLL models
- NNLS -> NNLR in documentation
- Minor typos


desilinguist:
Can we change the `Model fit` table so that the number of responses and number of features are integers instead of floats?

rsmtool/input.py
```diff
@@ -381,25 +381,54 @@ def process_json_fields(json_obj):
...
+                try:
+                    new_json_obj[field] = json.loads(lower_case_value)
+                except ValueError:
+                    raise ValueError(error_message)
+
+                # confirm that the new value is a boolean since
+                # json will also parse an integer as integer
+                if not type(new_json_obj[field]) == bool:
+                    raise ValueError(error_message)
     return new_json_obj
```
desilinguist: Wouldn't it be simpler to just check whether the value matches a regex, e.g., 

```
re.match(r'^(true|false)$', a, re.I)
```

PR #19, opened 2016-06-20
[RSMR-264] Making report printing a bit better

desilinguist:
- Reset margins and padding.
- Hide table of contents and python packages section.
- Hide `Out[]` prompts.
- Make highlighting work in printed reports.


aloukina:
This looks good. Do we also want to fix the printing for RSMCompare or not at this time?

PR #24, opened 2016-07-06
Fix License to actually declare copyright

jkahn:
Discovered missing when comparing to output of the `lice` tool.

part of JOSS review:
https://github.com/openjournals/joss-reviews/issues/33


desilinguist:
Thanks Jeremy!

PR #34, opened 2016-07-24
Suppress expected warnings in nose output.

desilinguist:
Use the `catch_warnings()` context manager to suppress expected warnings in nose output.


aloukina:
This is ready to be merged.

PR #49, opened 2016-08-01
Allow transformations leading to inf/NaN for rsmpredict

aloukina:
Previously if old model specified a transformation that would create inf/NaN values when applied to new data (e.g. sqrt(-1), rsmpredict would raise an error and terminate.
The new behavior is to simply exclude such responses in the same way as we are already excluding the responses with non-numeric feature values and to show a warning.


rsmtool/preprocess.py
```diff
@@ -299,13 +305,19 @@ def apply_inverse_transform(name, data, sd_multiplier=4):
     ValueError
         If the transform is applied to a feature that can
         be zero or to a feature that can have different
-        signs.
+        signs and raise_error is set to 'Trie'
```
desilinguist: typo "Trie".

rsmtool/preprocess.py
```diff
@@ -349,20 +370,26 @@ def apply_sqrt_transform(name, data):
     ------
     ValueError
         If the transform is applied to a feature
-        that has negative values.
+        that has negative values and raise_error is set to true.
```
desilinguist: raise_error should be in double backquotes?

rsmtool/preprocess.py
```diff
@@ -383,26 +413,40 @@ def apply_log_transform(name, data):
     ------
     ValueError
         If the transform is applied to a feature that
-        can be zero or negative.
+        can be zero or negative and raise_error is set to true.
```
desilinguist: Same here.

rsmtool/preprocess.py
```diff
@@ -548,6 +616,9 @@ def preprocess_feature(data,
     exclude_zero_sd : bool, optional
         Check `data` has a zero
         std. dev. Defaults to False.
+    raise_transformation_error : bool, optional
+        Raise error if any of the transformations lead to inf values
```
desilinguist: inf should probably also be quoted (``Inf``?)

rsmtool/preprocess.py
```diff
@@ -305,7 +305,7 @@ def apply_inverse_transform(name, data,
     ValueError
         If the transform is applied to a feature that can
         be zero or to a feature that can have different
-        signs and raise_error is set to 'Trie'
+        signs and `raise_error` is set to 'True'
```
desilinguist: You don't need to quote the True since you don't in other places but it's fine.

PR #53, opened 2016-08-11
Feature/check for id for rsmcompare

aloukina:
1. Since RSMCompare now allows for missing input files, I added extra checks to make sure the paths/ids specified by the user actually point to existing experiments.
       a. Added test that the tool raises an error if a wrong experiment directory is specified.
       b. Added check for wrong experiment_id and a relevant test.

2. RSMCompare now reports the correlation in raw feature values between old and new version.

2. Fixed the bug which forced confusion matrix to name columns starting with 1 as well as the bug which incorrectly handled different score levels in score distribution table (both rsmcompare).

3.  Started re-factoring rsmcompare code to make it easier to test.


rsmtool/format_utils.py
```diff
@@ -0,0 +1,106 @@
+"""
```
desilinguist: Why can't we just put this in `rsmtool.utils`?

PR #55, opened 2016-08-25
Add highlight color css code and css in comparison header

bndgyawali:


rsmtool/notebooks/comparison/header.ipynb
```diff
@@ -285,7 +305,13 @@
     "        series.append(df.apply(summary_func))\n",
     "    res = pd.concat(series, axis=1)\n",
     "    res.columns = ['MEAN', 'SD', 'MEDIAN', 'MIN', 'MAX']\n",
-    "    return res"
+    "    return res\n",
+    "\n",
+    "def color_highlighter(x, low=0, high=1, prec=3, absolute=False):\n",
+    "    abs_x = abs(x) if absolute else x\n",
```
aloukina: I suppose we could in future move that functions into some kind of utilities module too.

rsmtool/notebooks/header.ipynb
```diff
@@ -61,10 +61,8 @@
...
+    "javascript_path = os.environ.get(\"JAVASCRIPT_PATH\")\n",
```
aloukina: Perhaps it would be best to keep all os.environ.get commands together? This way if somebody wants to set them manually, they are all in the same place?

desilinguist: Yes, I agree.

rsmtool/notebooks/comparison/header.ipynb
aloukina: @bndgyawali  if you decide to do that for this PR, there is a whole family of formatting functions (like int_or_float_format), so it would make sense to move all of them.

bndgyawali: At least this is the one needed now. If you would want me to do that, I could move these all into a separate python program and load them in both headers.

desilinguist: Let's do that later.

aloukina: The other ones already exist in duplicate copies in both headers.

desilinguist:
I also updated `environment.yml` for read the docs since that suffered from the same issue. I tested the build and it works fine.

PR #64, opened 2016-09-12
Fix the loading of Javascript sorting library in header notebooks.

desilinguist:
- Use `notebook_path` in `report.py` when constructing new paths.
- Explicitly read `javascript/sort.js` in both header notebooks to get around weird encoding issues in RSMApp.


aloukina:
Looks good!

PR #72, opened 2016-10-24
Bugfix: error in pca

aloukina:
- PCA had a built-in expectation that the number of components will always be greater or equal than N features. If this was not the case, the code would crash.


rsmtool/analysis.py
```diff
@@ -345,7 +345,7 @@ def compute_pca(df, selected_features):
 
     df_components = pd.DataFrame(pca.components_)
     df_components.columns = selected_features
-    df_components.index = ['PC{}'.format(i) for i in range(1, len(selected_features) + 1)]
+    df_components.index = ['PC{}'.format(i) for i in range(1, len(df_components.index) + 1)]
```
desilinguist: Is `len(df_components)` different than `len(df_components.index)` here?

rsmtool/analysis.py
```diff
@@ -358,7 +358,7 @@ def compute_pca(df, selected_features):
     df_variance = df_variance[['Eigenvalues', 'Percentage of variance', 'Cumulative percentage of variance']]
 
     # set the row names and take the transpose
-    df_variance.index = ['PC{}'.format(i) for i in range(1, len(selected_features) + 1)]
+    df_variance.index = ['PC{}'.format(i) for i in range(1, len(df_variance.index) + 1)]
```
desilinguist: And same here?

tests/test_analysis.py
```diff
@@ -91,3 +92,13 @@ def test_that_metrics_helper_works_for_data_with_the_same_label():
...
         evals = metrics_helper(same_human_scores, system_scores)
         assert_equal(evals.isnull().values.sum(), 1)
+
+
+def test_compute_pca_less_components_than_features():
+    # test pca when we have less components than features
+    df = pd.DataFrame({'a':range(100)})
+    for i in range(100):
+        df[i] = df['a']*i
+    (components, variance) = compute_pca(df, df.columns)
```
aloukina: Hm.. May be I should an actual test statement here, not just implicitly testing it does not crash...

PR #82, opened 2017-02-17
Fix plot positions and improve subgroup plots

desilinguist:
- With the last release of RSMTool (v5.5.1), we had upgraded to matplotlib v2.0 and that has seemed to cause some problems with plots in the reports appearing out of place. This branch fixes that issue.
- In addition, this branch also improves the subgroup plots. If there are 6 or fewer subgroups, the plots are smaller and two per row. For larger number of subgroups, the current model of a large one-per-row plot is followed. In both cases, the font sizes are a little larger for easier reading.


rsmtool/notebooks/header.ipynb
```diff
@@ -39,7 +39,6 @@
     "\n",
     "from functools import partial\n",
     "from os.path import abspath, exists, join\n",
-    "from textwrap import wrap\n",
```
aloukina: You are sure we are not using it?

desilinguist: It got moved to `rsmtool.utils` since it's used in the new function I added: `get_subgroup_plot_params()`. I also tested reports after removing this to make sure nothing is broken.

rsmtool/notebooks/features_by_group.ipynb
```diff
@@ -38,7 +38,7 @@
     "    df_train_combined.reset_index(drop=True, inplace=True)\n",
     "\n",
     "    # decide on the the height per plot\n",
-    "    height_per_plot = 2 if len(features_used) > 15 else 4\n",
```
aloukina: We are not longer applying this condition, right? (I don't think it's part of the function either). Does it still look fine with 50 features?

desilinguist: The condition is in the function. In the old version of this notebook, `height_per_plot` is used to compute the total height of the figure by multiplying with the number of features. That is what [these](https://github.com/EducationalTestingService/rsmtool/blob/bugfix/plot-positions/rsmtool/utils.py#L423) [two](https://github.com/EducationalTestingService/rsmtool/blob/bugfix/plot-positions/rsmtool/utils.py#L428) lines do in the new function.

aloukina: Ok, the report looks good for a large N features.

PR #87, opened 2017-03-03
Better handling of unused features

aloukina:
- The features that have not been used in the model are now excluded from the datasets before they are sent to SKLL for prediction. This makes the prediction step much faster on large datasets.
- Removed the warning which is no longer necessary
- A minor change to the utility script update_skll_model.py to allow it to be used with other tools.


tests/update_skll_models.py
```diff
@@ -15,11 +15,12 @@
 
 import glob
 
-from os import remove
+from os import getcwd, remove
 from os.path import join, dirname
 from skll import Learner
 
-TEST_DIR = dirname(__file__)
+#TEST_DIR = dirname(__file__)
```
desilinguist: Do we need this commented line?

aloukina: No :)

desilinguist: :-p

PR #88, opened 2017-03-14
Correct treatment of ID columns with leading zeros

aloukina:
* Added converters to the notebook header to allow correct treatment of candidate IDs with leading zeros
* Modified the test utility functions to catch discrepancies caused by missing leading zero.

These changes do not have any implications on RSMTool functionality but missing converters in the header could cause issues in special/custom notebooks.


rsmtool/notebooks/header.ipynb
```diff
@@ -202,80 +202,84 @@
    "outputs": [],
    "source": [
     "# Read in the training and testing features, both raw and pre-processed\n",
-    "# Make sure that the `spkitemid` column is read as a string\n",
+    "# Make sure that the `spkitemid` and `candidate` columns are read as strings \n",
+    "# to preserve any leading zeros\n",
     "# We filter DtypeWarnings that pop up mostly in very large files\n",
     "\n",
+    "string_columns = ['spkitemid', 'candidate']\n",
+    "converter_dict = dict([(column, str) for column in string_columns if column])\n",
```
desilinguist: You can use a dictionary comprehension which will be much more readable

`{column: str for column in string_columns}`

rsmtool/test_utils.py
```diff
@@ -147,8 +147,20 @@ def check_csv_output(csv1, csv2):
...
 
     """
-    df1 = pd.read_csv(csv1, index_col=0)
-    df2 = pd.read_csv(csv2, index_col=0)
+
+    # make sure that the main id columns are read as strings since
+    # this may affect merging in custom notebooks
+    string_columns = ['spkitemid', 'candidate']
+
+    converter_dict = dict([(column, str) for column in string_columns if column])
```
desilinguist: can use dictionary comprehension here too.

rsmtool/test_utils.py
```diff
@@ -157,10 +169,10 @@ def check_csv_output(csv1, csv2):
     # convert any integer columns to floats in either data frame
     for df in [df1, df2]:
         for c in df.columns:
-            if df[c].dtype == np.int64:
+            if not c in string_columns and df[c].dtype == np.int64:
```
desilinguist: if you are reading them as strings to begin with, then do we need this test at all?

PR #93, opened 2017-04-03
Fix bug in SMD computation

desilinguist:
The denominator for SMDs should be using population standard deviations, not the ones computed over the subgroups.


desilinguist:
I still need to add a note about raw scores and SMDs to the report and also to the documentation.

PR #98, opened 2017-10-02
Display evaluation by subgroup as table.

Lguyogiro:
Displays the `Evaluation by <subgroup>` data as a table in addition to the graphs in the `rsmeval` report.


desilinguist:
Hmm, why did the coverage go down?

Lguyogiro:
My last pushed commit, `add float formatting to table display`, caused Warnings to be output, there was an increase in coverage for that build (#253), because line 218 in `test_utils.py`, which increments the `report_warnings` counter. The most recent commit fixed that issue,so line 218  is no longer reached and there is an apparent decrease in coverage.

desilinguist:
Got it. Thanks! LGTM 👍

PR #100, opened 2017-10-03
Feature/provide subgroup eval as table

Lguyogiro:
This is an addition to PR #98 . The subgroup table is now sortable, and highlights SMD values outside of the threshold range [-0.1, 0.1].


rsmtool/notebooks/evaluation_by_group.ipynb
```diff
@@ -27,14 +27,23 @@
     "                 ('RMSE', 'trim'): []}\n",
     "\n",
     "colprefix = 'scale' if use_scaled_predictions else 'raw'\n",
-    "metrics = dict([('{}.{}_{}'.format(k[0], colprefix, k[1]), v) for k,v in basic_metrics.items()])\n",
+    "metrics = dict([('{}.{}_{}'.format(k[0], colprefix, k[1]), v)\n",
+    "                for k,v in basic_metrics.items()])\n",
     "num_metrics = len(metrics)\n",
+    "formatters = {'{}.{}_{}'.format(k[0], colprefix, k[1]): partial(color_highlighter, low=min(v), high=max(v))\n",
```
desilinguist: I am not sure why you need this?

Lguyogiro:
I meant to include that line `formatters = {'{}.{}_{}'.format(k[0], colprefix, k[1]): partial(color_highlighter, low=min(v), high=max(v))` so that if we ever alter the thresholds, it is just a matter of doing it once in the `basic_metrics` dictionary. I've updated that part now.

rsmtool/notebooks/evaluation_by_group.ipynb
Lguyogiro: I meant to include that line formatters = {'{}.{}_{}'.format(k[0], colprefix, k[1]): partial(color_highlighter, low=min(v), high=max(v)) so that if we ever alter the thresholds, it is just a matter of doing it once in the basic_metrics dictionary. I've updated that part now.

desilinguist: I see what you are trying to do now but it doesn't seem very readable to me. Here's how I would suggest doing it:

```python
formatter_dict = {}
for (smd_variant, [smd_low, smd_high]) in [(k, v) for k, v in metrics.items() if k.startswith('SMD')]:
    formatter_dict[smd_variant] = partial(color_highlighter, low=smd_low, high=smd_high)
```

And then you can pass `formatter_dict` as the value to the `formatters` keyword argument in `to_html()`.

PR #103, opened 2018-01-04
Upgrade Python, SKLL and other dependencies

desilinguist:
- Upgrade python to v3.6.
- Upgrade SKLL to v1.5.
- Upgrade pandas to v0.22.0. This requires changing some of our code since some of the pandas functionality we currently use has been deprecated (`.ix[]`, `sortlevel()`, `get_value()`).
- In v0.19.0 of scikit-learn,  the implementation of `explained_variance_` in their PCA module underwent some bugfixes. Due to this, the `pcavar.csv` files had to be updated for all tests.
- Had to add new filters for a `FutureWarning` since the latest `statsmodels` uses some deprecated code from `pandas`. 
- Because of the change in scikit-learn version, all `rsmpredict` experiment tests yielded warnings about models trained with the older scikit-learn version. So, I changed all of the model files to be the new ones.
- Because of the change in models, a couple of other files for a specific `rsmpredict` test also had to be updated (`lr-predict-illegal-transformation`). Particular attention should be paid to this change to make sure that it was actually necessary.


tests/data/experiments/lr-predict-illegal-transformations/output/predictions.csv
```diff
@@ -1,4 +1,4 @@
 spkitemid,raw,sc1,scale,raw_trim,raw_trim_round,scale_trim,scale_trim_round
-RESPONSE_1,-8.638996331151485,4,-13.535417043079162,0.50002,1,0.50002,1
```
aloukina: Do we know why this happened?

tests/data/experiments/lr-predict-illegal-transformations/existing_experiment/output/lr_postprocessing_params.csv
```diff
@@ -1,2 +1,2 @@
 h1_mean,h1_sd,train_predictions_mean,train_predictions_sd,trim_max,trim_min
-3.42,0.8154323146156515,3.420000000000002,0.5799500693653682,6.0,1.0
```
aloukina: Oh, you changed the parameters of the existing experiment, right?

desilinguist: Yeah, the file changed values because of some changes in pandas.

desilinguist:
@aloukina and I talked and it looks like the parameters and predictions files changed most likely because we must have made changes to the data in the past. Also, in the past, @aloukina used a script that basically read the model and wrote it back to disk (such that it would then become compatible with the latest version of sklearn) which is why we didn't run into changed prediction files before.

As a check, I will also check in the coefficients CSV files into the `existing_experiment/output` directory such that in the future, we will know if the models have changed. Stay tuned for a commit.

PR #108, opened 2018-01-09
Minor changes for upgrade

aloukina:
Changed the script for updating skll model files to account for new logging structure in SKLL. This is an auxiliary script used for maintenance.


tests/update_skll_models.py
```diff
@@ -33,6 +33,7 @@ def update_model(model_file):
 
     # now load the SKLL model
     model = Learner.from_file(model_file)
+    model.logger = ''
```
desilinguist: Can we add a comment here that we won’t need this once we update SKLL to v1.5.1? I will also file an issue so that we remember.

desilinguist:
I am going to merge this without waiting for the build since it won’t affect any of the tests anyway.

Issue #116, opened 2018-01-18
Add config option allowing users to save output in other formats (e.g. Excel)


jbiggsets:
The refactored version of RSMTool includes a `Writer` class with a `write_experiment_output()` method. This method can optionally take a `file_format` argument, which allows output files to be saved as Excel, CSV, or JSON.

Given that we have this functionality, we may want to add a configuration option (and associated checks in the `ConfigurationParser` class) that allows users to specify other output formats for their experiments. 

I think we can add this in after the refactoring is finished.

PR #127, opened 2018-01-24
Allow users to save output in different file formats

jbiggsets:
This PR is intended to do the following.

- Addresses issue #116.
    - Updates configuration parameters to include a `file_format` argument, which allows users to specify CSV, TSV, or XLSX output.
    - Adds a new environment variable (`FILE_FORMAT`) that is set for the `rsmtool`, `rsmpredict`, and `rsmeval` contexts.
    - Swaps out `pd.read_csv()` with `DataReader.read_from_file()` in most notebooks.
    - Replaces file names hardcoded with a `.csv` extension (in notebooks and elsewhere) with the correct file extension (using the string's `.format()` method).
    - Adds a function to automatically determine the output directory file extensions for `rsmsummarize` and `rsmcompare` experiments.
    - Adds new tests for TSV and XLSX outputs across all relevant contexts.
    - Incorporates miscellaneous updates to scripts and modules to ensure that all of the above work properly.

- Changes `nbconvert` timeout from `3600` to `4800` to try to mitigate [this](https://github.com/jupyter/jupyter_client/issues/154) issue.

Note that I did **not** make any updates to the documentation at this stage. I will save all documentation updates for a separate PR.


rsmtool/utils.py
```diff
@@ -586,6 +595,56 @@ def compute_subgroup_plot_params(group_names, num_plots):
...
+    Raises
+    ------
+    ValueError
+        If any files in the directory have different extensions,
+        and are in the list of possible output extensions.
+    """
+    extension = 'csv'
+    extensions_identified = [os.path.splitext(name)[-1].replace('.', '')
+                             for name in os.listdir(directory)
+                             if any(name.endswith(ext) for ext in POSSIBLE_EXTENSIONS)]
```
desilinguist: This seems a little complicated and inefficient? How about a different approach:

```python
def has_files_with_extension(dir, ext):
    files_with_extension = glob(join(dir, f'*.{ext}'))
    return len(files_with_extension) > 0

extensions_identified = [ext for ext in extensions if has_files_with_extension(directory, ext)]
```

rsmtool/reporter.py
```diff
@@ -313,7 +313,7 @@ def convert_ipynb_to_html(notebook_file, html_file):
 
         # set a high timeout for datasets with a large number of features
         report_config = Config({'ExecutePreprocessor': {'enabled': True,
-                                                        'timeout': 3600},
+                                                        'timeout': 4800},
```
desilinguist: Since this didn't seem to make a difference, I'd revert this change.

rsmtool/test_utils.py
```diff
@@ -138,7 +139,7 @@ def do_run_summary(source, config_file):
     run_summary(config_file, experiment_dir)
 
 
-def check_csv_output(csv1, csv2):
+def check_csv_output(csv1, csv2, file_format='csv'):
```
desilinguist: Should this really be called `check_csv_output` if `csv` is only one possible extension?

rsmtool/test_utils.py
```diff
@@ -264,7 +281,7 @@ def check_scaled_coefficients(source, experiment_id):
                        check_less_precise=True)
 
 
-def check_all_csv_exist(csv_files, experiment_id, model_source):
+def check_all_csv_exist(csv_files, experiment_id, model_source, file_format='csv'):
```
desilinguist: Same here.

rsmtool/test_utils.py
```diff
@@ -277,30 +294,33 @@ def check_all_csv_exist(csv_files, experiment_id, model_source):
...
-                          "_pca.csv",
-                          "_pcavar.csv",
-                          "_pcor_score_all_data.csv",
-                          "_pred_processed.csv",
-                          "_pred_train.csv",
-                          "_score_dist.csv",
-                          "_train_preprocessed_features.csv",
-                          "_test_preprocessed_features.csv",
-                          "_postprocessing_params.csv"
+    csv_must_have_both = ["_confMatrix.{}".format(file_format),
```
desilinguist: And here.

rsmtool/test_utils.py
```diff
@@ -324,9 +344,12 @@ def check_consistency_files_exist(csv_files, experiment_id):
         List of CSV files generated by a test.
     experiment_id : str
         The experiment ID.
+    file_format : str, optional
+        The format of the output files.
+        Defaults to 'csv'.
     """
-    csv_must_have = ["_consistency.csv",
-                     "_degradation.csv"]
+    csv_must_have = ["_consistency.{}".format(file_format),
```
desilinguist: And here.

rsmtool/comparer.py
```diff
@@ -253,54 +256,63 @@ def load_rsmtool_output(self, csvdir, figdir, experiment_id, prefix, groups_eval
             A dictionary with experiment figures.
         """
 
+        extension = get_output_directory_extension(csvdir, experiment_id)
+
         csvs = defaultdict(pd.DataFrame)
```
desilinguist: Should this really be called `csvs`?

rsmtool/test_utils.py
jbiggsets: Yeah, I was debating this. :)

Changing them is okay with me. Suggestions on naming convention? Should we just replace "csv" with "file" or is there something that makes more sense?

desilinguist: `check_generated_output`?

rsmtool/comparer.py
```diff
@@ -223,16 +226,16 @@ def process_confusion_matrix(conf_matrix):
         conf_matrix_renamed.columns = ['human {}'.format(x) for x in conf_matrix.columns]
         return conf_matrix_renamed
 
-    def load_rsmtool_output(self, csvdir, figdir, experiment_id, prefix, groups_eval):
+    def load_rsmtool_output(self, filedir, figdir, experiment_id, prefix, groups_eval):
```
aloukina: I'd go with `outputdir` but this is more of nitpick :)

rsmtool/comparer.py
```diff
@@ -245,63 +248,72 @@ def load_rsmtool_output(self, csvdir, figdir, experiment_id, prefix, groups_eval
 
         Returns
         -------
-        csvs : dict
-            A dictionary with ``.csv`` outputs converted to pandas data
+        files : dict
```
aloukina: and `outputs` but see above :)

rsmtool/modeler.py
```diff
@@ -865,7 +866,13 @@ def train_score_weighted_lr(self, df_train, feature_columns):
 
         return learner, fit, df_coef, used_features
 
-    def train_builtin_model(self, model_name, df_train, experiment_id, csvdir, figdir):
+    def train_builtin_model(self,
+                            model_name,
+                            df_train,
+                            experiment_id,
+                            csvdir,
```
aloukina: then this should also be `filedir` or whatever we end up naming it

rsmtool/modeler.py
```diff
@@ -961,18 +975,28 @@ def train_builtin_model(self, model_name, df_train, experiment_id, csvdir, figdi
...
         self.learner = learner
 
         return learner
 
-    def train_skll_model(self, model_name, df_train, experiment_id, csvdir, figdir):
+    def train_skll_model(self,
+                         model_name,
+                         df_train,
+                         experiment_id,
+                         csvdir,
```
aloukina: and here...

tests/test_utils.py
```diff
@@ -224,3 +226,27 @@ def test_compute_subgroups_with_wrapping_and_five_plots():
...
     eq_(expected_subgroup_plot_params, subgroup_plot_params)
+
+
+def test_has_files_with_extension_true():
+    directory = 'tests/data/files'
+    result = has_files_with_extension(directory, 'csv')
+    eq_(result, True)
+
+
+def test_has_files_with_extension_flase():
```
aloukina: "false"?

rsmtool/utils.py
```diff
@@ -626,9 +648,8 @@ def get_output_directory_extension(directory, experiment_id):
         and are in the list of possible output extensions.
     """
     extension = 'csv'
-    extensions_identified = [os.path.splitext(name)[-1].replace('.', '')
-                             for name in os.listdir(directory)
-                             if any(name.endswith(ext) for ext in POSSIBLE_EXTENSIONS)]
+    extensions_identified = [ext for ext in POSSIBLE_EXTENSIONS
+                             if has_files_with_extension(directory, ext)]
 
     extensions = set(extensions_identified)
```
desilinguist: Actually, if you need a set, just use a set comprehension (`{ext for ext in ...}`) instead of a list comprehension above.

jbiggsets: Good call. :)

I'll make that change and fix a couple of the minor things Anastassia mentioned.

rsmtool/utils.py
```diff
@@ -648,20 +648,19 @@ def get_output_directory_extension(directory, experiment_id):
         and are in the list of possible output extensions.
     """
     extension = 'csv'
-    extensions_identified = [ext for ext in POSSIBLE_EXTENSIONS
-                             if has_files_with_extension(directory, ext)]
+    extensions_identified = set(ext for ext in POSSIBLE_EXTENSIONS
```
desilinguist: Sorry, you can just use `{ .. }` instead of `[ .. ]` and that just gives you a set comprehension. I thought I had mentioned that in my earlier comment.

PR #135, opened 2018-02-02
Allow objectives for skll models

desilinguist:
- Add an optional `skll_objective` option to the RSMTool configuration file, with a default of `None`.
  - Raise an exception if specified objective is not a valid SKLL objective. 
  - Raise a warning if the objective is specified for a non-SKLL learner.
- Modify `Modeler.train()` and `Modeler.train_skll_model()` to use the specified custom objective if specified but otherwise use `neg_mean_squared_error`.
- Pass objective to notebooks and show in SKLL model notebook.
- Remove stray `ipdb` trace left over from a previous commit. 
- Clean up imports and spacing in test files.
- Add new tests: one for invalid an SKLL objective and two experiment tests for `LinearSVR` and `Ridge` with custom objectives. 
- Add `skll_objective` to RSMTool docs and fix an already existing issue with the placement of `file_format`.


tests/test_preprocessor.py
```diff
@@ -644,8 +644,6 @@ def test_filter_on_flag_column_nothing_left(self):
         flag_dict = {'flag1': [1, 0, 14], 'flag2': ['TD']}
 
         df_new, df_excluded = self.fpp.filter_on_flag_columns(bad_df, flag_dict)
```
jbiggsets: We could probably remove the assignment here, too, since `filter_on_flag_columns()` should raise a `ValueError`.

desilinguist: Which assignment?

jbiggsets: `df_new` and `df_excluded` are assigned but never used.

desilinguist: Ah, I see. Pretty sure `flake8` will catch that so let's take care of that when we are doing the code style review.

Issue #113, opened 2018-01-18
Option to load images as clickable thumbnails in HTML report


jbiggsets:
Large HTML reports with lots of images can take a very long time to load. It would be helpful to allow users to generate clickable thumbnail images (rather than full-sized images) in the report. The images can be saved in the `figures` folder.

In addition, it may be useful to produce a zipped version of the experiment.

desilinguist:
I guess the second part of this (creating a zip file) will be addressed in a separate PR?

desilinguist:
Addressed by #142.

PR #142, opened 2018-02-06
Generate Clickable Thumbnails

jbiggsets:
This PR is to address issue #113.

- Add `use_thumbnails` configuration parameter. 
- Adds the `show_thumbnail` function to `utils.py` to convert images to clickable thumbnails.
- Update notebooks to allow optional generation of clickable thumbnails, rather than images.
- Update `comparer.py` so that the `Comparer` class returns paths to image files, rather than base64-encoded image strings.
- Add various tests to ensure that `show_thumbnail` is working properly.
- Add note to in notebooks to indicate whether thumbnails are being displayed.


rsmtool/utils.py
```diff
@@ -61,6 +63,7 @@
             'use_scaled_predictions_new': False,
             'select_transformations': False,
             'standardize_features': True,
+            'images_to_thumbnails': False,
```
desilinguist: Why not just `use_thumbnails`? For boolean variables, it's usually best to have a verb in there just like we do for most of the others.

rsmtool/utils.py
```diff
@@ -667,6 +674,120 @@ def get_output_directory_extension(directory, experiment_id):
     return extension
 
 
+class Thumbnail:
```
desilinguist: The actual technique for generating the thumbnails works really well but I have some suggestions for the semantics of this class:

1. I would actually modify the  `__init__()` method to take the path to the image file and then do the checks that the image exists, get its relative path, and then store the path in the `image_path` attribute (along with setting `current_id` to 0). This makes it much clearer when creating a `Thumbnail` instance in user land. 

2. I would rename `convert_to_html()` to just be `to_html()` to follow other libraries' conventions.

3. I would rename `convert` to `show()`.

4. Right now, there's no indication that the thumbnail is clickable. I would add `cursor: pointer` to the CSS and also add something like "click to enlarge" text below the thumbnail. It would also be nice to have some thing in the top of the report in bold saying that the images have been converted to thumbnails and the user can click on them to enlarge them. We may be sending these reports around to people who aren't that tech savvy.

rsmtool/notebooks/header.ipynb
```diff
@@ -152,6 +154,16 @@
...
    ]
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# initialize thumbnail converter object\n",
+    "thumbnail = Thumbnail()"
```
desilinguist: Ah, I think I have misunderstood how you use `Thumbnail`. You are essentially using it as a singleton just instantiating it once and then using it to convert any new file. So, an instance of `Thumbnail` is not really per image file. It's just a converter that keeps track of IDs - that's a little strange way to do it but I guess it's more efficient than having multiple Thumbnail instances around. However, in this case, we need to call it something else, perhaps `ThumbnailConverter` and make it clear how we use it.

jbiggsets: Sure, I'll rename it and address the comments below. My logic for using this pattern was twofold:

(1) I think it's more efficient, as you said, and (2) it ensures that `ids` are tracked and incremented, and therefore always unique.

rsmtool/utils.py
jbiggsets: This all makes sense except for the `__init__()` method, which can't take the file path, since I'm only instantiating the object once. I can check that the file exists in the `to_html()` method. Let me know if that sounds all right.

doc/config_rsmtool.rst
```diff
@@ -185,6 +185,17 @@ standardize_features *(Optional)*
 If this option is set to ``false`` features will not be standardized by dividing by the mean and multiplying by the standard deviation. Defaults to ``true``.
 
 
+.. _use_thumbnails_rsmtool:
```
aloukina: Perhaps move this next to other report-related fields (section lists)?

jbiggsets: Sure, I can rearrange these in my `specify-features-as-list` branch, since I still need to update the documentation there.

desilinguist:
@jbiggsets I think you missed this comment in my review: " It would also be nice to have some thing in the top of the report in bold saying that the images have been converted to thumbnails and the user can click on them to enlarge them. We may be sending these reports around to people who aren't that tech savvy."

desilinguist:
So, I ran the `lr-with-custom-sections-and-order` test manually after adding `use_thumbnails` to it. I am seeing that x-axis labels in the the length correlations plot (and thumbnail) seem to be truncated. However, I don't see this truncation of the labels if I turn off the thumbnails. See the two attached images below:

With thumbnails:
screen shot 2018-02-06 at 5 06 05 pm

Without thumbnails:
screen shot 2018-02-06 at 5 04 43 pm

jbiggsets:
The problem seems to be that the figure is actually getting saved this way (whether or not we convert to thumbnails). In the other figure plots where the axis labels are horizontal, I called `plt.tight_layout()`, and that seems to solve the problem. I'll update, and also add the notes at the top in `header.py` saying the images have been converted to thumbnails.

One thing that's worth pointing out is that the x-axes being cut off seems to be an **existing** issue with `rsmtool`. Even if I use the current `master` branch and look at the `svg` output, the axis is cut off. This doesn't happen when you call `plt.show()`, so it looks correct in the HTML report, but not in the saved image file.

desilinguist:
Hmm, I didn't know that. Let's fix this in this release then. Let me know if the `tight_layout()` call doesn't fix this.

jbiggsets:
@desilinguist OK, calling `plt.tight_layout()` solved the problem for me. I didn't see any other images where this was happening, but let me know if you noticed anything else.

Issue #128, opened 2018-01-24
Add new learners added to the latest SKLL release


aloukina:
We'll need to add the to: 
- List of learners
- Documentation
- Tests (note that for models based on optimization we only test that the test actually runs since the predictions might differ across systems)

desilinguist:
Addressed by #144.

PR #144, opened 2018-02-06
Add new SKLL learners and tests.

desilinguist:
- Add new regressors from SKLL v1.5.
- Add two new tests to cover two of the new regressors.
- Update `skll_model` notebook to show coefficients for the newly added linear regressors.

Addresses #128.


doc/config_rsmtool.rst
```diff
@@ -84,7 +84,9 @@ The name for the column containing the human scores in the training data. If set
 
 .. note::
 
-    All responses with non-numeric values in either ``train_label_column`` or ``test_label_column`` and/or those with non-numeric values for relevant features will be automatically excluded from model training and evaluation. By default, zero scores in either ``train_label_column`` or ``test_label_column`` will also be excluded. See :ref:`exclude_zero_scores_rsmtool` if you want to keep responses with zero scores.
+    1. All responses with non-numeric values in either ``train_label_column`` or ``test_label_column`` and/or those with non-numeric values for relevant features will be automatically excluded from model training and evaluation. By default, zero scores in either ``train_label_column`` or ``test_label_column`` will also be excluded. See :ref:`exclude_zero_scores_rsmtool` if you want to keep responses with zero scores.
+
+    2. If ``train_label_column`` and ``test_label_column`` contain floating point scores, they may be rounded for some of the evaluation methods, e.g. when computing kappa and the confusion matrices.
```
aloukina: "May be rounded for some" is too vague in my opinion. If you do have floating point scores, you want to know what is going on. "They will be rounded for the following evaluations: ....".
Are we sure we are not rounding them for correlations too?

rsmtool/notebooks/evaluation.ipynb
```diff
@@ -71,7 +71,9 @@
    "source": [
     "#### Association statistics\n",
     "\n",
-    "The table shows the standard association metrics between human scores and machine scores. Note that for the computation of kappas scores are always rounded.\n"
+    "The table shows the standard association metrics between human scores and machine scores. \n",
+    "\n",
+    "*Please note that if your data contained floating point scores, they are rounded for the computation of `kappa` and `wtkappa`.*\n"
```
aloukina: And agreement?

rsmtool/utils.py
```diff
@@ -37,16 +37,24 @@
...
+               'BayesianRidge',
                'DecisionTreeRegressor',
+               'DummyRegressor',
                'ElasticNet',
                'GradientBoostingRegressor',
+               'HuberRegressor',
                'KNeighborsRegressor',
+               'Lars',
                'Lasso',
+               'LinearRegression',
```
aloukina: If we expose SKLL Linear Regression, shall it be added to `skll_linear_models` above?

tests/test_experiment_rsmtool.py
```diff
@@ -120,6 +120,122 @@ def test_run_experiment_lr_subset_feature_file_and_feature_file():
...
+        expected_csv_file = join(expected_output_dir, csv_filename)
+
+        if exists(expected_csv_file):
+            yield check_file_output, csv_file, expected_csv_file
+
+    yield check_generated_output, csv_files, experiment_id, 'skll'
+    yield check_report, html_report
+
+
+def test_run_experiment_dummyregressor_custom_objective():
```
aloukina: This should technically throw a warning that the objective will be ignored, right?

tests/test_experiment_rsmtool.py
```diff
@@ -120,6 +120,122 @@ def test_run_experiment_lr_subset_feature_file_and_feature_file():
...
+        expected_csv_file = join(expected_output_dir, csv_filename)
+
+        if exists(expected_csv_file):
+            yield check_file_output, csv_file, expected_csv_file
+
+    yield check_generated_output, csv_files, experiment_id, 'skll'
+    yield check_report, html_report
+
+
+def test_run_experiment_dummyregressor():
```
aloukina: We don't add predictions/model parameters to tests for most SKLL learners because of their non-deterministic nature. But this is not really concern for dummy regressor so we might as well add them.

tests/test_experiment_rsmtool.py
desilinguist: No, that's only for non-SKLL learner and this is technically in SKLL.

doc/config_rsmtool.rst
desilinguist: Can you make the change you think is appropriate?

tests/test_experiment_rsmtool.py
desilinguist: Good point. I will add those.

rsmtool/notebooks/evaluation.ipynb
desilinguist: Again, I will defer to you about what you want to say here.

rsmtool/utils.py
desilinguist: Actually, it doesn't make any sense to expose this since we already have `LinearRegression` in our built in models list. I will remove this.

doc/config_rsmtool.rst
aloukina: Re-open the issue and remove it from this PR? I think it's more then just wording - it's about checking where we round the scores in the code too.

desilinguist: The issue is still open. I will edit the PR description.

tests/data/experiments/dummyregressor/output/DummyRegressor_eval.csv
```diff
@@ -0,0 +1,7 @@
+,N,h_mean,h_sd,h_min,h_max,sys_mean,sys_sd,sys_min,sys_max,corr,wtkappa,R2,kappa,exact_agr,adj_agr,SMD,RMSE
```
aloukina: Nice to see that the same score does not break the eval table :)

rsmtool/notebooks/skll_model.ipynb
```diff
@@ -47,7 +47,7 @@
    },
    "outputs": [],
    "source": [
-    "skll_linear_models = ['Ridge', 'LinearSVR', 'Lasso']"
+    "skll_linear_models = ['Ridge', 'BayesianRidge', 'LinearSVR', 'Lasso', 'Lars', 'HuberRegressor', 'TheilSenRegressor']"
```
aloukina: We are sure that this will work, right?

desilinguist: Yes, I tested it locally :)

PR #150, opened 2018-02-07
Bugfix/#137 wrong file path

aloukina:
* Added error capture/error message when configuration file specifies path to a non-existing file
* Added an extra check to the DataReader to make sure the paths passed to the reader are not None (Nones usually emerge when `locate_file` was not able to locate the file)

### DIFF
--- a/rsmtool/reader.py
+++ b/rsmtool/reader.py
@@ -65,8 +65,9 @@ class DataReader:
         assert len(filepaths) == len(framenames)
 
         # Make sure that there are no Nones in the filepaths
-        frames_with_no_path = [framenames[i] for i in range(len(framenames)) if filepaths[i] is None]
-        if len(frames_with_no_path) > 0:
+        
+        if None in filepaths:
+            frames_with_no_path = [framenames[i] for i in range(len(framenames)) if filepaths[i] is None]
             raise ValueError("No path specified for {}".format(' ,'.join(frames_with_no_path)))
 
         # Assign names and paths lists
--- a/rsmtool/reader.py
+++ b/rsmtool/reader.py
@@ -54,6 +54,8 @@ class DataReader:
             If any elements in file_converters are not dict.
         NameError
             If file converter name does not exist in the dataset.
+        ValueError
+            If filepath for a given file is None
         """
 
         # Default datasets list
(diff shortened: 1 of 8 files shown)


rsmtool/reader.py
```diff
@@ -62,6 +64,11 @@ def __init__(self,
         # Make sure filepaths length matches frame names length
         assert len(filepaths) == len(framenames)
 
+        # Make sure that there are no Nones in the filepaths
+        frames_with_no_path = [framenames[i] for i in range(len(framenames)) if filepaths[i] is None]
+        if len(frames_with_no_path) > 0:
+            raise ValueError("No path specified for {}".format(' ,'.join(frames_with_no_path)))
+
```
jbiggsets: This is really nit-picky, but I wonder if it's slightly more efficient to just check whether `None` is in the `filepaths` list first, and if so, only then create the `frames_with_no_path` list and raise the `ValueError`. 

```
if None in filepaths:
    frames_with_no_path = [name for idx, name in enumerate(framenames) if filepaths[idx] is None]
    raise ValueError("No path specified for {}".format(' ,'.join(frames_with_no_path)))
```

(This is what you did in `rsmtool.py` below.)

rsmtool/rsmtool.py
```diff
@@ -97,13 +97,27 @@ def run_experiment(config_file_or_obj,
...
+    if None in file_paths:
+        indices_with_no_paths = [i for i in range(len(file_paths))
+                                 if file_paths[i] is None]
+
+        error_message_list = ["{} file {} not found.".format(file_names[i],
+                                                             file_paths_org[i])
+                              for i in indices_with_no_paths]
+
+        raise FileNotFoundError('\n'.join(error_message_list))
+
```
jbiggsets: Do we need this check here if we're going to get a similar `ValueError` when we initialize the `DataReader` in line 129? Is this just pre-empting that?

aloukina: This will give a more specific error message saying what file is missing at what path. Unfortunately, we don't pass the original path to the DataReader to return a meaningful message and I find them very useful.

desilinguist: I am not a fan of the way this exception looks for multiple files. Why not do something like:

```python
    if None in file_paths:
        indices_with_no_paths = [i for i in range(len(file_paths))
                                 if file_paths[i] is None]
        missing_file_paths = [file_paths_org[i] for i in indices_with_no_paths]
        raise FileNotFoundError('The following files were not found: {}'.format(repr(missing_file_paths)))
```

aloukina: This change would obscure what was the destination of the wrong path (e.g. 'train' or 'test' or 'feature'). I suppose we can try dropping it and see if it gets annoying

Issue #126, opened 2018-01-19
Links to .csv files from .html report


jbiggsets:
It could be useful to have links in .html report which would allow you to immediately open the specific .csv files for this given analysis rather than asking people to remember our sometimes not very transparent naming scheme. I was just trying to show all of this to Jay and it's a bit messy even with output_csv.md file.  The problem is of course that if the report is forwarded without the output folder the links will be lost. 

 A perhaps better alternative could be to generate a separate .html report which contains only explanations+links to available .csv files and store this report under output. Then somebody from IDEAS who ran the tool and then wants to play around with .csv could quickly locate them by using this "report".

desilinguist:
I think we can easily have links next to the sections/tables where the CSV files are used.

desilinguist:
Addressed by #153.

PR #153, opened 2018-02-08
Add notebooks with hyper-links to intermediate files

jbiggsets:
This PR addresses issue #126 by adding new notebooks to the `rsmtool`/`rsmeval`, `rsmsummarize`, and `rsmcompare` contexts, which display links to intermediate files generated during the experiment. This update required the following changes:

- Adding a utility function to generate hyper-linked lists of intermediate files.
- Adding new unit tests for this function.
- Making minor updates to a handful of existing functional test configuration files.
- Adding new notebooks with links to intermediate files for the above-mentioned contexts.
- Making minor updates to the code in `reporter.py` and `comparer.py` to get this working.
- Updating configuration the documentation.

**Update (2/25)**:
- Removed the file paths notebook in the `rsmcompare` context.
- Updated the file paths notebook in the `rsmsummarize` context to show _only_ intermediate files generated by `rsmsummarize`.
- Made other minor, miscellaneous updates to address @aloukina's PR comments.


doc/config_rsmsummarize.rst
```diff
@@ -43,6 +43,8 @@ RSMTool provides pre-defined sections for ``rsmsummarize`` (listed below) and, b
 
     - ``evaluation``: Compares the standard set of evaluations recommended for scoring models on the evaluation data.
 
+    - ``intermediate_file_paths_summarize``: Shows links to all of the intermediate files that were generated while running the summary.
```
desilinguist: Why do we need the `_summarize` bit?

doc/config_rsmcompare.rst
```diff
@@ -100,6 +100,8 @@ RSMTool provides pre-defined sections for ``rsmcompare`` (listed below) and, by
 
     - ``notes``: Notes explaining the terminology used in comparison reports.
 
+    - ``intermediate_file_paths_compare``: Shows links to all of the intermediate files that were generated while running the comparison.
```
desilinguist: And why do we need the `_compare` bit here?

rsmtool/utils.py
```diff
@@ -856,12 +856,86 @@ def show_thumbnail(path_to_image, image_id):
...
+    replace_dict : dict, optional
+        A dictionary which makes file names to descriptions.
+        Defaults to empty dictionary.
+
+    Yields
+    ------
+    html_string : str
+        HTML list item with file description and link.
+    """
+    html_string = ("""<li><b>{}</b>: <a href="{}" download>{}</a></li>""")
```
desilinguist: Why initialize this each time if it's not changing? Can't we define it as a constant up top?

rsmtool/utils.py
```diff
@@ -856,12 +856,86 @@ def show_thumbnail(path_to_image, image_id):
...
+    files = [f for f in os.listdir(output_dir) if f.endswith(file_format)]
+
+    for file in sorted(files):
+        relative_file = os.path.relpath(os.path.join(output_dir, file))
```
desilinguist: Can we use `pathlib` throughput this function? And also, we should probably look up just the prefix in the replacement dictionary rather than the whole filename? So, just `betas` instead of `lr_betas.csv`? And we could make the code a bit more readable:

```python
files = Path(output_dir).glob('*.{}'.format(file_format))

for file in sorted(files):
    relative_file = file.relative_to(os.getcwd())
    relative_name = relative_file.stem.replace(experiment_id, '')

    if relative_name in replace_dict:
        descriptive_name = replace_dict[relative_name]
    else:
        descriptive_name_components = relative_name.split('_') 
        descriptive_name = ' '.join(descriptive_name_components).strip().title()

   yield ...
```

desilinguist:
So, I tried this out locally and the second gets really, really long. May be in a future version, we can try to reduce the spacing and perhaps group files into table cells? Can you please file an issue @jbiggsets?

Also, I am not sure "References to Intermediate Files" is very clear. How about "Links to Intermediate Files"? And also, a little text below the title saying something like: "Click on the hyperlinks below to see the intermediate experiment files output as part of this experiment" or something like that?

rsmtool/comparer.py
```diff
@@ -329,9 +328,9 @@ def load_rsmtool_output(self, filedir, figdir, experiment_id, prefix, groups_eva
         # read in disattenuated correlations by group
         for group in groups_eval:
             group_dis_corr_file = join(filedir,
-                                     '{}_disattenuated_correlations_by_{}.{}'.format(experiment_id,
-                                                                                     group,
-                                                                                     extension))
+                                       '{}_disattenuated_correlations_by_{}.{}'.format(experiment_id,
+                                                                                       group,
+                                                                                       extension))
```
aloukina: It's a complete nitpick, but why do we use `extension` rather than `file_format` (like elsewhere in the code)?

rsmtool/notebooks/comparison/intermediate_file_paths.ipynb
```diff
@@ -0,0 +1,68 @@
...
+   "source": [
+    "## Links to Intermediate Files"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Click on the hyperlinks below to see the intermediate experiment files output as part of this comparison."
+   ]
```
aloukina: My biggest concern here is that when the users share that report they probably won't be sharing the original directories or if they do the paths would change when they zip everything up. I'd probably just keep the skeleton for future comparison outputs and not share the paths to the original files.

rsmtool/notebooks/summary/intermediate_file_paths.ipynb
```diff
@@ -0,0 +1,63 @@
...
+   "metadata": {},
+   "source": [
+    "## Links to Intermediate Files"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Click on the hyperlinks below to see the intermediate experiment files output as part of this summary."
```
aloukina: "generated as part of this summary" ?

rsmtool/notebooks/comparison/intermediate_file_paths.ipynb
aloukina: Actually, the links in the current reports generated by test (I am looking at `lr-self-compare-different-formats`) do not work so it will be a problem.

rsmtool/notebooks/summary/intermediate_file_paths.ipynb
```diff
@@ -0,0 +1,63 @@
...
+   "source": [
+    "from rsmtool.utils import show_files"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
```
aloukina: Same issue as for `rsmcompare`: current paths in reports do not work and I am almost certain in most cases we'll be generating broken links here as the users will move things around.

aloukina: On the other hand, `rsmsummarize` does generate it's own outputs that we want to list here.

rsmtool/notebooks/comparison/intermediate_file_paths.ipynb
jbiggsets: Ah, I see the problem. The relative paths aren't working properly for the `rsmcompare` and `rsmsummary`.I can fix that, but should we just get rid of the links instead? We can change the title to "List of intermediate files" or something?

aloukina: I would vote for commenting out the notebook for `RSMCompare` for the time being (or just keeping it empty) since there are no outputs generated (we'll want to add it back when we address the outputs issue so please don't delete all the great code you already wrote!). 

 For `RSMSummarize` we can just give the links to `rsmsummarize` outputs in the `output` folder - hopefully this is a straightforward change.

What do you think?

desilinguist:
I agree with @aloukina.

rsmtool/notebooks/comparison/intermediate_file_paths.ipynb
jbiggsets: Yup! Makes sense!

Issue #129, opened 2018-01-24
Allow using SKLL classifiers with probability output


aloukina:

desilinguist:
I am not really sure what this means. Is this to allow the user to use expected values as predictions?

aloukina:
Yes, to give user access to the same approach as you used in BEA paper: "Therefore, for all four classifiers, we use the predict_proba() method of their scikit-learn implementations to obtain probability distributions over the possible score points and then compute the expected value using the distribution as the final classifier prediction."

desilinguist:
Addressed by #156.

PR #156, opened 2018-02-09
Improve SKLL integration with RSMTool

desilinguist:
This PR improves integration of RSMTool with SKLL in addition to addressing issue #129.

#### Code

1. Remove static list of SKLL learners and use `importlib` to check any given model name against the `skll.learner` module directly in a new function called `is_skll_model()`. This means that any and all SKLL learners (classifiers and regressors) are now available to RSMTool. 
2. For consistency, also create a new function called `is_builtin_model()`. 
3. Allow computation of expected scores (probability-weighted averages over score points) from probabilistic SKLL classifiers for both `rsmtool` and `rsmpredict`.
   - Add a new `predict_expected_scores` option to the configuration files for `rsmtool` and `rsmpredict`, which is `False` by default. 
    - Add checks to configuration parser to quickly raise an error if the specified model name for RSMTool experiments are not probabilistic classifiers. The same check is used to raise an error for built-in models since they are not classifiers.
    - Modify `Modeler.train_skll_model()` to train with probabilities enabled if we want expected scores. 
    - Modify `Modeler.predict()` to take in the min and max scores and the `predict_expected_scores` flag.
   - Add a new general purpose function called `compute_expected_scores_from_model()` which takes in a SKLL model, a `FeatureSet` object, the min and max scores, and computes expected scores using probability distributions obtained from the model. It also explicitly checks for errors and raises exceptions. 
    - Call this function from both `rsmtool` and `rsmpredict` in the appropriate places. 
4. Have `train_skll_model()` choose `f1_score_micro` as the default objective if the SKLL learner is a classifier. Have it also return the objective which we can then pass into the report and then not have to worry about checking the objective in the notebooks. 
5. Improve calls to `train_skll_model()` and `train_builtin_model()` to use keyword arguments correctly. 
6. Modify log messages to show "expected scores" when generating predictions.
7. Add a note to the top of the report if the experiment used expected scores.
8. Some other minor changes for readability and style.

#### Tests

Added several new tests.
- Tests for the general function `compute_expected_scores_from_model()`.
- Tests for the configuration parsing piece.
- Several new experiment tests for classification, classification with custom objectives, and classification with expected scores for both rsmtool as well as rsmpredict experiments.
- Tests for rsmpredict to catch various errors with expected scores. 

#### Documentation

- Add documentation about expected scores for both `rsmtool` and `rsmpredict` and add explicit notes about the expected scores potentially being different on different machines due to small probability values.
- Move `skll_objective` back down to optional fields since we want all 4 required `rsmtool` argument to be together. Added links from the `model` field to `skll_objective` and `predict_expected_scores` fields.


desilinguist:
Something occurred to me today - I wonder if we can make the expected scores more deterministic if we round the probabilities before taking the dot product with the scores. I will try that.

desilinguist:
@jbiggsets @aloukina You can already start reviewing this PR when you get a chance.

Issue #159, opened 2018-02-13
Update `DataWriter()` to allow dictionary or `DataContainer` to be passed


jbiggsets:
It might be useful in some cases to just pass a dictionary to the `DataWriter()` `write_experiment_output()` method instead of a `DataContainer()`. This should be an easy update.

desilinguist:
Addressed by #160.

PR #160, opened 2018-02-13
Allow dictionary to to be passed to `DataWriter` `write_experiment_output()` method

jbiggsets:
In some cases, it may be useful to pass a dictionary to the `DataWriter` `write_experiment_output()` method, rather than having to instantiate a `DataContainer` object. 

This PR addresses issue #159. It also incorporates a new test to check that a `dict` can be passed to `write_experiment_output()`.

### DIFF
--- a/tests/test_writer.py
+++ b/tests/test_writer.py
@@ -52,6 +52,44 @@ class TestDataWriter:
         assert_frame_equal(container.dataset1, ds_1_csv)
         assert_frame_equal(container.dataset1, ds_1_xls)
 
+    def test_dictionary_save_files(self):
+
+        data_sets = {'dataset1': pd.DataFrame(np.random.normal(size=(100, 2)),
+                                              columns=['A', 'B']),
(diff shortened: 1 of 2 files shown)


tests/test_writer.py
```diff
@@ -45,14 +45,51 @@ def test_data_container_save_files(self):
...
         assert_frame_equal(container.dataset1, ds_1_xls)
 
+    def test_dictionary_save_files(self):
+
+        data_sets = {'dataset1': pd.DataFrame(np.random.normal(size=(100, 2)),
+                                              columns=['A', 'B']),
+                     'dataset2': pd.DataFrame(np.random.normal(size=(120, 3)),
+                                              columns=['A', 'B', 'C'])}
+
+        directory = 'temp_directory'
```
aloukina: Can we rename this into something more random so that we don't accidentally delete an existing `temp_directory` when running this test? This is actually the case for all tests in this script.

jbiggsets: Sure.

PR #170, opened 2018-02-17
Refactor `update_files.py`

desilinguist:
- Refactor all of the core file updating code into a class called `FileUpdater` located in `rsmtool.test_utils.py`. This makes the code much more organized and also makes it much easier to write tests for it in a subsequent release. 
- The code now doesn't raise any exceptions. It just records all of the information when then `run()` method is called and prints everything out when `print_report()` is called. 
- The code is also fully modular in that nothing is hardcoded. You need to pass in the directory containing the tests as well as the directory containing the updated test outputs. In addition, it also automatically gets all of the `test_experiment_*.py` files from the tests directory and uses those. This will make it possible to use the _same_ script for RSMExtra without any changes. 
- Tests decorated with `@raises` which do not need to be updated are now skipped much more directly rather than relying on an indirect signal previously. 
- Sometimes you might want to name the file `test_experiment_X.py` but _not_ update its test data automatically. For example, the file `test_experiment_rsmcompare.py` does actually contain experiment tests but since they do not contain an output directory, they do not need to be updated (for now). However, this file does get picked up by `update_files.py` because of its name and then we get reports of missing directories. I solved this by simply adding an `_AUTO_UPDATE` variable to the top of all of the `test_experiment_*.py` files and setting it to `True` for all files except for `test_experiment_rsmcompare.py`. `FileUpdater.run()` now checks for this before actually processing a file. 

In short, this change should allow us to use the same script for RSMExtra without any changes and also _always_ print out a useful report for the data update process.

### DIFF
--- a/rsmtool/test_utils.py
+++ b/rsmtool/test_utils.py
@@ -879,40 +879,39 @@ class FileUpdater(object):
                 continue
 
             # iterate over all the members and focus on only the experiment functions
-            # and try to get the source and the experiment ID since that's what we
-            # need to update the test files
+            # but skip over the functions that are decorated with '@raises' since those
+            # functions do not need any test data to be updated. For the rest, try to get
+            # the source since that's what we need to update the test files
             for member_name, member_object in getmembers(test_module):
                 if isfunction(member_object) and member_name.startswith('test_run_experiment'):
                     function = member_object
 
-                    # first we check if it's the parameterized function and if so
-                    # we can easily get the source and the experiment ID from
-                    # the parameter list
+                    # get the qualified name of the member function
+                    member_qualified_name = member_object.__qualname__
+
+                    # check if it has 'raises' in the qualified name and skip it, if it does
+                    if 'raises' in member_qualified_name:
+                        continue
+
+                    print(member_name)
+
+                    # otherwise first we check if it's the parameterized function and if so
+                    # we can easily get the source from the parameter list
                     if member_name.endswith('parameterized'):
                         for param in function.parameterized_input:
                             source_name = param.args[0]
                             skll = param.kwargs.get('skll', False)
                             self.update_source(source_name, skll=skll)
 
-                    # if it's another function, then we actually inspect the source
-                    # to get the source and experiment_id and this can never be a
-                    # SKLL experiment since those should always be run parameterized
+                    # if it's another function, then we actually inspect the code
+                    # to get the source. Note that this should never be a SKLL experiment
+                    # since those should always be run parameterized
                     else:
                         function_code_lines = getsourcelines(function)
-                        experiment_id_line = [line for line in function_code_lines[0]
-                                              if re.search(r'experiment_id = ', line)]
-
-                        # if there was no experiment ID specified, it was either a
-                        # a decorated test function (which means it wouldn't have an 'output'
-                        # directory anyway) or it was a compare or prediction test function
-                        # which are not a problem for various reasons.
-                        if experiment_id_line:
-
-                            # get the name of the source directory
-                            source_line = [line for line in function_code_lines[0]
-                                           if re.search(r'source = ', line)]
-                            source_name = eval(source_line[0].strip().split(' = ')[1])
-                            self.update_source(source_name)
+                        source_line = [line for line in function_code_lines[0]
+                                       if re.search(r'source = ', line)]
+                        source_name = eval(source_line[0].strip().split(' = ')[1])
+                        self.update_source(source_name)
 
     def print_report(self):
         """
--- a/rsmtool/test_utils.py
+++ b/rsmtool/test_utils.py
@@ -872,7 +872,7 @@ class FileUpdater(object):
         # adapted from: http://stackoverflow.com/questions/67631/how-to-import-a-module-given-the-full-path
         for test_suffix in self.test_suffixes:
             test_module_path = join(self.tests_directory, 'test_experiment_{}.py'.format(test_suffix))
-            test_module = SourceFileLoader('test_experiment', test_module_path).load_module()
+            test_module = SourceFileLoader('loaded_{}'.format(test_suffix), test_module_path).load_module()
 
             # make sure that the module tells us that it wants to be updated
             if not test_module._AUTO_UPDATE:
--- a/rsmtool/test_utils.py
+++ b/rsmtool/test_utils.py
@@ -884,9 +884,10 @@ class FileUpdater(object):
             test_module_path = join(self.tests_directory, 'test_experiment_{}.py'.format(test_suffix))
             test_module = SourceFileLoader('loaded_{}'.format(test_suffix), test_module_path).load_module()
 
-            # make sure that the module tells us that it wants to be updated
-            if not test_module._AUTO_UPDATE:
-                continue
+            # skip the module if it tells us that it doesn't want the data for its tests updated
+            if hasattr(test_module, '_AUTO_UPDATE'):
+                if not test_module._AUTO_UPDATE:
+                    continue
 
             # iterate over all the members and focus on only the experiment functions
             # but skip over the functions that are decorated with '@raises' since those
--- a/rsmtool/test_utils.py
+++ b/rsmtool/test_utils.py
@@ -731,6 +731,9 @@ class FileUpdater(object):
         experiment tests.
     deleted_files : list
         List of files deleted from `tests directory`.
+    updated_files : list
+        List of files that have either (really) changed in the updated outputs
+        or been added in those outputs.
     missing_or_empty_sources : list
         List of source names whose corresponding directories are either
         missing under `updated_outputs_directory` or do exist but are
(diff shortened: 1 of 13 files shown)


desilinguist:
The coverage went down because it's including `rsmtool/test_utils.py` in the coverage which we don't really test at all right now. I will tell Travis to ignore it for now.

rsmtool/test_utils.py
```diff
@@ -700,3 +706,242 @@ def check_subgroup_outputs(output_dir, experiment_id, subgroups, file_format='cs
...
+            "test_experiment_" and located in the `tests_directory` to find
+            the tests that are to be updated.
+        updated_outputs_directory (TYPE): Description
+        deleted_files (list): List of files deleted from `tests directory`.
+        updated_files (list): List of files added or updated for the tests
+            under `tests_directory`.
+        missing_or_empty_sources (list): List of source names whose corresponding
+            directories are either missing under `updated_outputs_directory` or
+            do exist but are empty.
+    """
```
jbiggsets: Can we change these to `numpy`-style docstrings, rather than Google-style?

```
Attributes
----------
tests_directory : str
    Path to the directory containing the tests whose
    outputs are to be updated.
```

desilinguist: Weird, I could have sworn I had set my `AutoDocstring` to numpy and not google since it works fine for methods. I will redo.

rsmtool/test_utils.py
```diff
@@ -700,3 +706,242 @@ def check_subgroup_outputs(output_dir, experiment_id, subgroups, file_format='cs
...
+        suffix = Path(filename).suffix
+        return suffix == '.model' or \
+            suffix == '.npy'or \
+            stem.endswith('_postprocessing_params') or \
+            stem.endswith('_eval') or \
+            stem.endswith('_eval_short') or \
+            stem.endswith('_confMatrix') or \
+            stem.endswith('_pred_train') or \
+            stem.endswith('_pred_processed') or \
+            stem.endswith('_score_dist')
```
jbiggsets: This is probably not a big deal, but would we ever be adding to the list of possible `skll` suffixes or stems here? Is it easier to put these into lists, rather than having so many `or` statements?

```
possible_suffixes = ['.model', '.npy']
possible_stems = ['_postprocessing_params', '_eval', '_eval_short', ...]

return any(suffix == s for s in possible_suffixes) or \
    any(stem.endswith(s) for s in possible_stems)
```

rsmtool/test_utils.py
```diff
@@ -700,3 +706,242 @@ def check_subgroup_outputs(output_dir, experiment_id, subgroups, file_format='cs
...
+        # since those are likely old files from old versions that we do not need
+        existing_output_only_files = dir_comparison.right_only
+        for file in existing_output_only_files:
+            remove(existing_output_path / file)
+
+        # Next find all the NEW files in the updated outputs but do not include
+        # the config JSON files or the OLS summary files or
+        # the evaluation/prediction/model files for SKLL models
+        new_files = dir_comparison.left_only
+        new_files = [f for f in new_files if not f.endswith('_rsmtool.json') and not f.endswith('_rsmeval.json') and not f.endswith('_ols_summary.txt')]
```
jbiggsets: Again, would it be better to just use `any()` here and a list of filename suffixes, in case we expand on this in the future?

rsmtool/test_utils.py
desilinguist: Ah, it was set to `auto_google`, which means it tries to autodetect the style based on the other docstrings but falls back to google style if it can't.

tests/update_files.py
```diff
@@ -157,90 +57,25 @@ def main():
     elif run_test_suite != 'y':
         print('Invalid answer. Exiting.')
         sys.exit(1)
+    else:
+        print()
+
+    # iterate over the given tests directory and find all files named
+    # `test_experiment_*.py` and get their suffixes for use with the
+    # FileUpdater object.
+    suffixes = [re.sub(r'test_experiment_', '', p.stem)
```
jbiggsets: Is it really necessary to have the `_AUTO_UPDATE` global variables in each test module. Can't we just ignore `rsmcompare` here?

rsmtool/test_utils.py
```diff
@@ -700,3 +706,242 @@ def check_subgroup_outputs(output_dir, experiment_id, subgroups, file_format='cs
...
+                    function = member_object
+
+                    # get the qualified name of the member function
+                    member_qualified_name = member_object.__qualname__
+
+                    # check if it has 'raises' in the qualified name and skip it, if it does
+                    if 'raises' in member_qualified_name:
+                        continue
+
+                    print(member_name)
```
jbiggsets: Do we need this print here?

tests/update_files.py
desilinguist: It goes back to your other comment for the other things: what if we want to add more modules in the future we don't want to update? I am happy to hardcode it if you guys prefer.

rsmtool/test_utils.py
desilinguist: No, we don't :)

tests/update_files.py
jbiggsets: No, that makes sense. 

Is there somewhere we can document the we have to add `_AUTO_UPDATE` at the top if/when we add new `test_experiment_*.py` files? I'm just worried I'll totally forget about it. :)

desilinguist: Yes, it will be in the testing document that I will be writing.

rsmtool/test_utils.py
```diff
@@ -700,3 +706,251 @@ def check_subgroup_outputs(output_dir, experiment_id, subgroups, file_format='cs
...
+
+        # if the directory for this source does not exist on the updated output
+        # side, then that's a problem and something we should report on later
+        try:
+            assert updated_output_path.exists()
+        except AssertionError:
+            self.missing_or_empty_sources.append(source)
+            return
+
+        # if the updated output path does not exist, then create it
```
aloukina: updated -> existing

rsmtool/test_utils.py
```diff
@@ -700,3 +706,251 @@ def check_subgroup_outputs(output_dir, experiment_id, subgroups, file_format='cs
...
+                except AssertionError:
+                    pass
+                else:
+                    include_file = False
+
+            if include_file:
+                really_changed_files.append(changed_file)
+
+        # Copy over the new files as well as the really changed files
+        new_or_changed_files = new_files + really_changed_files
```
aloukina: Why don't we  then have `self.updated_files` and `self.new_files`?

tests/update_files.py
```diff
@@ -28,126 +29,25 @@
...
+from rsmtool.test_utils import FileUpdater
 
 
 def main():
     # set up an argument parser
     parser = argparse.ArgumentParser(prog='update_test_files.py')
-    parser.add_argument('outputs_dir', help="The path to the directory containing the updated test outputs")
+    parser.add_argument('--tests',
+                        dest='tests_dir',
+                        help="The path to the RSMTool tests directory")
```
aloukina: 'existing'?

tests/update_files.py
```diff
@@ -28,126 +29,25 @@
...
+    parser.add_argument('--outputs',
+                        dest='outputs_dir',
+                        help="The path to the directory containing the test "
+                             "outputs (usually test_outputs)")
```
aloukina: 'updated'?

tests/update_files.py
aloukina: Will you also add to the documentation that the test script should be named `test_experiment_*`?

desilinguist: Yes it’s in there already.

rsmtool/test_utils.py
desilinguist: You mean ALSO have `self.new_files` In addition to `self.updated_files`?

tests/update_files.py
```diff
@@ -28,126 +29,25 @@
...
-    # return lists of files we deleted and updated
-    return sorted(existing_output_only_files), sorted(new_or_changed_files)
+from rsmtool.test_utils import FileUpdater
 
 
 def main():
     # set up an argument parser
     parser = argparse.ArgumentParser(prog='update_test_files.py')
-    parser.add_argument('outputs_dir', help="The path to the directory containing the updated test outputs")
+    parser.add_argument('--tests',
```
aloukina: Potentially a nitpick: these are treated as optional arguments when you call the script so you get a Traceback error (Path is None) if you forget to specify them.

rsmtool/test_utils.py
aloukina: Yes: `self.new_files = new files` and `self.updated_files = `really_changed_files`. I don't feel too strongly about it though.

desilinguist: That's a good idea. I can do that.

tests/update_files.py
desilinguist: I can just add `required=True` to these.

rsmtool/test_utils.py
desilinguist: Actually, I'd rather not do that since that requires a lot of changes all over the code.

aloukina: That's fine - I'll file an issue for when we feel like it in future

PR #178, opened 2018-02-23
Get pip working

desilinguist:
To test this in the absence of RSMTool PyPI packages, you can do the following:

```bash
conda create -n rsmpip python=3.6
source activate rsmpip
pip install -e .
MPLBACKEND=Agg <any RSMTool command line>
```


desilinguist:
Yes that’s correct since they are tied to master.

Issue #145, opened 2018-02-06
Force length column to be numeric


aloukina:
After we filter out flagged response, we should force `length_column` to be numeric before checking for missing values/std==0.

desilinguist:
Addressed by #182.

PR #182, opened 2018-02-25
Force length column to be numeric

jbiggsets:
This is a PR to address issue #145.

Please review, and make sure I have fully addressed the issue.


rsmtool/preprocessor.py
```diff
@@ -1440,6 +1440,14 @@ def filter_data(self,
                            "for the following features: "
                            "{}".format(', '.join(missing_features)))
 
+        # if ``length_column`` exists and is non-numeric, convert it to numeric;
+        # values that cannot be coerced to numeric will be set to ``np.nan``
+        if length_column and not np.issubdtype(df_filtered['length'], np.number):
+            df_filtered['length'] = pd.to_numeric(df_filtered['length'], errors='coerce')
+            logging.info("The `length` column is non-numeric, and is being converted "
```
jbiggsets: Should this be a warning?

rsmtool/preprocessor.py
```diff
@@ -1440,6 +1440,14 @@ def filter_data(self,
                            "for the following features: "
                            "{}".format(', '.join(missing_features)))
 
+        # if ``length_column`` exists and is non-numeric, convert it to numeric;
+        # values that cannot be coerced to numeric will be set to ``np.nan``
+        if length_column and not np.issubdtype(df_filtered['length'], np.number):
```
desilinguist: Can't we just do `all(np.isreal(df_filtered['length']))`?

jbiggsets: Hmm. It looks like this actually won't work due to [this](https://github.com/pandas-dev/pandas/issues/10766) issue.

```
In [1]: import numpy as np

In [2]: import pandas as pd

In [3]: df = pd.DataFrame({'col': [1, 2, 3, 'A', '1.0', 5]})

In [4]: np.isreal(df.col)
Out[4]: array([ True,  True,  True,  True,  True,  True], dtype=bool)
```

aloukina: For the features I think we just do the conversion without raising the warning, perhaps we can do the same here? Would save us a condition :)

jbiggsets: That's OK with me if it's OK with @desilinguist. :)

desilinguist:
Works for me :)

rsmtool/preprocessor.py
desilinguist: Oh, that's so stupid numpy!

PR #188, opened 2018-02-28
Allow specification of the tests directory via an environment variable

desilinguist:
For package testing, we want to be able to run tests that are located outside of the package. This PR allows that.  To test that this works:

1. Check out `master` and `pip install -e .` that in a new conda environment (say `rsmenv`).
2. Check out this branch separately into a separate directory (say `$RSMNEWDIR`). Make some change to  `tests/data/experiments/lr/output/lr_betas.csv` but don't commit it.
3. Go to $HOME
4. Activate `rsmenv` and run `TESTDIR=$RSMNEWDIR nosetests --nologcapture $RSMNEWDIR/tests/test_experiment_rsmtool_1.py`. The `lr` experiment test should now fail. This tells us that the test data is now being used from the given `TESTDIR` instead of from the installed version of RSMTool in the environment.

### DIFF
--- a/tests/test_experiment_rsmcompare.py
+++ b/tests/test_experiment_rsmcompare.py
@@ -1,3 +1,5 @@
+import os
+
 from os.path import join
 
 from nose.tools import raises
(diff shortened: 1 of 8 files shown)


desilinguist:
This PR is right now the blocker for me to test the conda and PyPI packages so if you guys can test and review it sooner rather than later, that'd be great :)

Issue #219, opened 2018-07-11
Update `get_thumbnail_as_html()` to allow for a different thumbnail path


jbiggsets:
Currently, the `utils` function `get_thumbnail_as_html()` assumes that the path to the thumbnail image is the same as the path to the full-sized images. However, in some cases, we may want to display a different image when the thumbnail is clicked.

It would be useful to have an optional `path_to_thumbnail` argument. If this argument is not passed, then the `path_to_image` will be used for both.

PR #220, opened 2018-07-12
Update `get_thumbnail_as_html()`

jbiggsets:
This PR addresses #219. It updatse `get_thumbnail_as_html()` to add an optional `path_to_thumbnail` argument. There are cases where we may want to use separate thumbnail and full-sized images. If no `path_to_thumbnail` is passed, then the `path_to_image` argument will be used for both the thumbnail and full-sized image.


jbiggsets:
I'm not sure what's going on with Coveralls, but my updates should be covered by the test I added.  Even if that weren't true, a 6.7% decrease doesn't make sense. I'll have to take a closer look tomorrow.

Issue #217, opened 2018-07-11
Move `SMD` into a separate function.


jbiggsets:
Currently, SMD is only calculated in the `Analyzer. metrics_helper()` method. It would be useful to most this into a separate function so users can call it outside of RSMTool. This may be true for some other metrics as well.

jbiggsets:
Added in `RSMTool ehancements` project, and will be incorporated into the next release.

PR #221, opened 2018-07-12
Add separate SMD function

jbiggsets:
This PR addresses #217. It creats adds a separate `standardized_mean_difference()` function that can be called outside `rsmtool`.

**Note:** This PR should not be reviewed until after #220 is reviewed.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -270,16 +270,14 @@ def test_get_output_directory_extension_error():
 def test_standardized_mean_difference():
 
     # test SMD
-
-    expected = 1 / 3
-    smd = standardized_mean_difference(9, 8, 3, 3)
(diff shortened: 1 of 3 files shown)


aloukina:
Since we are likely to add another SMD function in a foreseeable future, can we call this `SMD_Williamson`?

tests/test_utils.py
```diff
@@ -266,6 +267,32 @@ def test_get_output_directory_extension_error():
     get_output_directory_extension(directory, 'id_1')
 
 
+def test_standardized_mean_difference():
+
+    # test SMD
+
```
aloukina: (Nitpick) replace 3 with 4 to avoid rounding in `eq_`?

PR #224, opened 2018-07-17
Incorporate Coveralls parallel webhook in Travis builds

desilinguist:
As described in this [issue](https://github.com/lemurheavy/coveralls-public/issues/1164), we should be using the parallel webhook since we are running multiple builds in parallel. This PR incorporates this webhook according to the instructions [here](https://docs.coveralls.io/parallel-build-webhook).


desilinguist:
This is no longer necessary as the adequate changes have been made in another PR.

Issue #212, opened 2018-07-11
Add helper methods to `Container`


jbiggsets:
Add new helper methods to `Container` class:

- `drop()`: Drop particular data frames by name.
- `rename()`: Rename particular data frames.
- `get_frames()`: Select partiuclar frames by prefix or suffix.

PR #226, opened 2018-07-23
Update `DataContainer` class

jbiggsets:
This PR addresses issues #212 by adding some helper methods to the `DataContainer` class. These methods allow users to`drop()` and `rename()` data frames in the `DataContainer`, and to select data frames using a specified prefix or suffix.


rsmtool/container.py
```diff
@@ -127,6 +129,46 @@ def get_frame(self, key, default=None):
             return default
         return self._dataframes[key]
 
+    def get_frames(self, prefix=None, suffix=None):
```
aloukina: Nice! I love this!

rsmtool/container.py
```diff
@@ -127,6 +129,46 @@ def get_frame(self, key, default=None):
...
+        """
+        Get all data frames in the container that have
+        a specified prefix and/or suffix.
```
aloukina: Add "Note that the selection is case-insensitive"

rsmtool/container.py
```diff
@@ -260,6 +302,50 @@ def __iter__(self):
...
+        Parameters
+        ----------
+        name : str
+            The name of the data frame to drop from the
+            container object.
+
+        Returns
+        -------
+        self
+        """
```
aloukina: Warning message if the frame is not there?

rsmtool/container.py
```diff
@@ -260,6 +302,50 @@ def __iter__(self):
...
+            in the container object.
+        new_name : str
+            The the new name for the data frame
+            in the container object.
+
+        Returns
+        -------
+        self
+        """
+        df = self._dataframes[name]
```
aloukina: Do we also need to update the `data_path` is that's specified in the container?

tests/test_container.py
```diff
@@ -0,0 +1,50 @@
...
+    def test_get_frames_by_prefix(self):
+
+        container = DataContainer([{'frame': pd.DataFrame(), 'name': 'test_two'},
+                                   {'frame': pd.DataFrame(), 'name': 'test_three'},
+                                   {'frame': pd.DataFrame(), 'name': 'exclude'}])
+
+        frames = container.get_frames(prefix='test')
+        eq_(sorted(list(frames.keys())), sorted(['test_two', 'test_three']))
+
+    def test_get_frames_by_suffix(self):
```
aloukina: Since it's a relatively low effort, could we add two more tests? One where there is no match and the test where there is both suffix and prefix match?

rsmtool/container.py
jbiggsets: Yes, we should do that.

rsmtool/container.py
```diff
@@ -260,21 +307,76 @@ def __iter__(self):
...
+            self._names.remove(name)
+            self._dataframes.pop(name)
+            self._data_paths.pop(name)
+        else:
+            warnings.warn('The name `{}` is not in the '
+                          'container. No data frames will '
+                          'be dropped.'.format(name))
+        return self
+
+    def rename(self, name, new_name):
```
desilinguist: Don't you want a warning here too? Basically don't we want a warning anytime we try to look for a frame that doesn't exist? Perhaps we could use an accessor method or something that has the warning?

Issue #214, opened 2018-07-11
Make updates to allow `Preprocessor` to use pre-specified min/max truncation values


jbiggsets:
Currently, `Preprocessor` truncates values that are +/- 4 standard deviations from the mean training set value for each feature. In some cases, however, users may want to provide pre-defined minimum and maximum truncation values, rather than calculating them from the training set provided.

PR #228, opened 2018-07-23
Update `Preprocessor` class

jbiggsets:
This PR addresses #214. It incorporates updates to the `Preprocessor` class to allow users to use pre-defined truncation values to remove `min` and `max` outliers, rather having these boundaries calculated from the training set.


doc/config_rsmtool.rst
```diff
@@ -207,6 +207,17 @@ If set to ``true``, certain evaluations (confusion matrices, score distributions
     All evaluation metrics (e.g., kappa and pearson correlation) are automatically computed for *both* scaled and raw scores.
 
 
+.. _use_truncations:
+
+use_truncations *(Optional)*
+"""""""""""""""""""""""""""""""""""
+If set to ``true``, use the ``min`` and ``max`` columns specified the ``features`` file to clamp outlier feature values. This is useful if users would like to clamp feature values based on some pre-defined boundaries, rather than having these boundaries calculated based on the training set. Defaults to ``false``.
```
desilinguist: `... specified IN ...`

doc/config_rsmtool.rst
```diff
@@ -207,6 +207,17 @@ If set to ``true``, certain evaluations (confusion matrices, score distributions
...
+
+.. note::
+
+    If ``use_truncations`` is passed, a ``features`` file _must_ be specified, and this file _must_ include ``min`` and ``max`` columns. If no ``feature`` file is specified or these columns are missing, an error will be raised.
```
desilinguist: `is set` instead of `is passed`?

doc/config_rsmtool.rst
```diff
@@ -207,6 +207,17 @@ If set to ``true``, certain evaluations (confusion matrices, score distributions
     All evaluation metrics (e.g., kappa and pearson correlation) are automatically computed for *both* scaled and raw scores.
 
 
+.. _use_truncations:
+
+use_truncations *(Optional)*
```
aloukina: `use_truncation_thresholds`  - this is to avoid giving impression that when set to False you won't use truncation at all.

Issue #211, opened 2018-06-08
Update documentation for flag_column_test


aloukina:
We need to make it clear in the documentation that this field supersedes `flag_column` and edit the logging text to make it reflect train or test.

Issue #193, opened 2018-03-02
Typo in warning in RSMPredict


aloukina:
"The following extraenous features will be ignored"

aloukina:
Resolved by above commit

PR #239, opened 2018-09-10
Typos in documentation

aloukina:
* Fixed few typos in documentation and messages (#193)
* Based on feedback, added further detail about the use of `flag_column_test` and updated the code to indicate partition when checking the flag column (#211). @jbiggsets please check this does not break any of your code.

### DIFF
--- a/tests/test_configuration_parser.py
+++ b/tests/test_configuration_parser.py
@@ -1,9 +1,11 @@
 import json
+import logging
 import os
 import tempfile
 import warnings
 import pandas as pd
 
+from io import StringIO
 from os.path import dirname, join
 from shutil import rmtree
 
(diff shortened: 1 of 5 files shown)


tests/test_configuration_parser.py
```diff
@@ -387,11 +394,31 @@ def test_check_flag_column_convert_to_list(self):
         flag_dict = config.check_flag_column()
         eq_(flag_dict, {"advisories": ['0']})
 
+
+    def test_check_flag_column_convert_to_list_test(self):
+        config = Configuration({"flag_column": {"advisories": "0"}})
+        flag_dict = config.check_flag_column(partition='test')
+        eq_(flag_dict, {"advisories": ['0']})
```
jbiggsets: I wonder if these tests should check the actual text of the warning, to make sure that we are getting the correct message for each `partition` type. If that makes sense to you, I can try to update the tests.

Issue #241, opened 2018-10-25
Generate `.environment.json` so users can easily re-run the notebook interactively


jbiggsets:
Currently, when a user runs an RSMTool experiment, it generates an `.ipynb` which is then run using `nbconvert` to generate an HTML report. The `.ipynb` relies on a series of environment variables that are set beforehand.

However, if the user wishes to go back and re-run this notebook interactively, these environment variables may no longer be set. It might be useful to save a (hidden) `.environment.json` file with any RSMTool experiment. This can then be used either in lieu of or in combination with the environment variables, so that users who wish to go back and run the notebook interactively can easily do so.

jbiggsets:
This was incorporated in the latest release.

PR #242, opened 2018-10-30
Notebooks read from JSON file instead of environment variables

jbiggsets:
This is to address #241. 

- Update `reporter.py` so that each `create_report()` method saves a `.environ.json` file.
- Update all of the `header.ipynb` notebooks to read from the `.environ.json` file.
- Add a couple new tests to make sure that the  `.environ.json` works as expected.

I spot-checked about a dozen notebooks, just to make sure there was noting unexpected, even though the tests are passing. Would be great if someone else could do some spot-checking, too.


desilinguist:
Oh and the PR should say JSON file instead of YAML right?

rsmtool/reporter.py
```diff
@@ -599,84 +599,72 @@ def create_report(self,
...
-        os.environ['PREDICT_EXPECTED_SCORES'] = '1' if predict_expected_scores else '0'
+        # if the features subset file is None, just use empty string
+        feature_subset_file = ('' if config.get('feature_subset_file') is None
+                               else config['feature_subset_file'])
+
+        # if the min items is None, just set it to zero
+        min_items = (0 if config['min_items_per_candidate'] is None
+                     else config['min_items_per_candidate'])
+
+        environ_config = {'EXPERIMENT_ID': config['experiment_id'],
```
aloukina: We are doing this funny thing where we first capitalize all field names in .json and then convert them to lower case again as we read in the values in header. It's ok to keep as is.

tests/test_experiment_rsmtool_1.py
```diff
@@ -43,6 +46,53 @@ def test_run_experiment_parameterized(*args, **kwargs):
...
+                       source,
+                       '{}.json'.format(experiment_id))
+    do_run_experiment(source, experiment_id, config_file)
+
+    report_ipynb = join('test_outputs', source, 'report', '{}_report.ipynb'.format(experiment_id))
+    report_html = join('test_outputs', source, 'report', '{}_report.ipynb'.format(experiment_id))
+
+    del os.environ['RSM_REPORT_DIR']
+
+    Reporter.convert_ipynb_to_html(report_ipynb, report_html)
```
aloukina: add `check_report(html_report)` to make sure there are no errors/warnings within the report?

aloukina:
Thank you for doing this!

PR #243, opened 2018-10-31
Fix boxplots in features by groups

jbiggsets:
This PR addresses a few issues.

1. The `tight_layout()` option in the `feature_descriptives` notebook fails if there are too many features. To resolve this, we add a try-except block and skip the `tight_layout()` option if it is going to throw an error.

2. There was a pre-existing bug in the `features_by_group` notebook in these two lines:

```
min_value = df_plot.mean() - 4 * df_plot.std()
max_value = df_plot.mean() + 4 * df_plot.std()
```

The `df_plot` data frame actually has two columns -- the subgroup column and the  feature column -- but we only want the mean and standard deviation for the feature column. We resolved this by selecting the feature column when we calculate the mean and standard deviation.

3. If there are a large number of features, then the `features_by_group` plots will be far too big, and the following error will be thrown:

```
ValueError: Image size of 1500x66000 pixels is too large. It must be less than 2^16 in each direction.
```

To avoid this, we simply skip this section if the number of features is greater than 30.

### DIFF
--- a/rsmtool/notebooks/feature_descriptives.ipynb
+++ b/rsmtool/notebooks/feature_descriptives.ipynb
@@ -88,7 +88,7 @@
     "    # we don't want the entire notebook to fail\n",
     "    try:\n",
     "        plt.tight_layout(h_pad=1.0)\n",
-    "    except:\n",
+    "    except ValueError:\n",
     "        pass\n",
     "\n",
     "    imgfile = join(figure_dir, '{}_outliers.svg'.format(experiment_id))\n",


rsmtool/notebooks/feature_descriptives.ipynb
```diff
@@ -84,7 +84,13 @@
     "    legend.legendHandles[0].set_color(colors[0])\n",
     "    legend.legendHandles[1].set_color(colors[1])\n",
     "\n",
-    "    plt.tight_layout(h_pad=1.0)\n",
+    "    # we want to try to force `tight_layout()`, but if this fails\n",
+    "    # we don't want the entire notebook to fail\n",
+    "    try:\n",
+    "        plt.tight_layout(h_pad=1.0)\n",
+    "    except:\n",
```
desilinguist: Can we specify the actual exception name here if we can? This kind of non-specific exception catching is generally frowned upon.

Issue #254, opened 2019-05-06
custom labels for rsmsummarize experiments


aoifecahill:
It would be nice to be able to override the way that the various experiments that get summarized by rsmsummarize are referenced in the table. Currently it uses the experiment_id of the original experiment. 

One proposal is to have a dictionary that has an rsmsummarize specific experiment_id for each experiment directory that is being summarized.

aloukina:
Addressed by #263

Issue #194, opened 2018-03-02
Save .json in `rsmsummarize` and `rsmcompare`


aloukina:
We now save `.json` for `rsmtool` and `rsmeval`. I am running some multi-step experiments where it would be useful to have `.json` saved for the other two tools for trouble-shooting purposes.

aloukina:
Addressed by #263

PR #263, opened 2019-06-20
Allow custom model names for rsmsummarize

aloukina:
[To be reviewed after  #261 ]
This PR addresses #254 

* RSMSummarize now allows using custom names to display for each experiment
* Updated tests and documentation
* Some minor edits to SMD documentation to make it clearer and to ensure it renders correctly on readthedocs


aloukina:
@jbiggsets @desilinguist This is the next PR to review. I merged in the current master.

doc/tutorial_rsmsummarize.rst
```diff
@@ -44,6 +44,7 @@ Let's take a look at the options in our configuration file.
 - **Line 2**: We provide the ``summary_id`` for the comparison. This will be used to generate the name of the final report. 
 - **Line 3**: We give a short description of this comparison experiment. This will be shown in the report.
 - **Line 4**: We also give the list of paths to the directories containing the outputs of the experiments we want to compare.
+- **Line 5**: Since all experiments we want to compare used the same experiment ids (``ASAP2``), we list the names that we want to use for each experiment in the summary report.
```
desilinguist:
```suggestion
- **Line 5**: Since we want to compare experiments that all used the same experiment id (``ASAP2``), we instead list the names that we want to use for each experiment in the summary report. 
```

rsmtool/configuration_parser.py
```diff
@@ -952,7 +950,14 @@ def validate_config(self, context='rsmtool', inplace=True):
                             "the automatically selected transformations "
                             "and signs.")
 
-        # 12. Clean up config dict to keep only context-specific fields
+        # 12. If we have `experiment_names`, check that the length of the list
+        # matches the list of experiment_dirs.
+        if context == 'rsmsummarize' and new_config['experiment_names']:
+            if len(new_config['experiment_names']) != len(new_config['experiment_dirs']):
+                raise ValueError("The total number of specified experiment names should be the same"
```
desilinguist:
```suggestion
                raise ValueError("The number of specified experiment names should be the same"
```

rsmtool/configuration_parser.py
```diff
@@ -952,7 +950,14 @@ def validate_config(self, context='rsmtool', inplace=True):
...
+                                 " as the total number of specified experiment dirs")
```
desilinguist:
```suggestion
                                 " as the number of specified experiment directories.")
```

rsmtool/configuration_parser.py
```diff
@@ -1055,8 +1060,8 @@ def normalize_validate_and_process_config(self, context='rsmtool'):
         self._check_config_is_loaded()
 
         self.normalize_config()
-        self.validate_config(context=context)
         self.process_config()
+        self.validate_config(context=context)
```
desilinguist: Why did we change the order of these two? Was it wrong before?

rsmtool/rsmcompare.py
```diff
@@ -104,6 +104,9 @@ def run_comparison(config_file_or_obj, output_dir):
         else:
             configpath = os.getcwd()
 
+    logger.info('Saving configuration file.')
+    configuration.save(output_dir)
```
desilinguist: Can you explain a bit about this? Was it missing before?

rsmtool/rsmsummarize.py
```diff
@@ -75,7 +77,15 @@ def check_experiment_dir(experiment_dir, configpath):
...
                                     "experiments.".format(full_path_experiment_dir))
 
-        return jsons
+        # Raise an error if the user specified a list of experiment names 
+        # but we found several .jsons in the same directory
+        if experiment_name and len(jsons) > 1:
+            raise ValueError("{} contains several experiment jsons "
+                             "To use custom experiment names "
+                             "You must have a separate "
+                             "directory for each experiment.")
```
desilinguist: This format string is missing the argument, right? Is the user supposed to know about experiment JSONs? Perhaps we should say something more general like:

> "{} seems to contain the output of multiple experiments. In order to use custom experiment names, you must have a separate directory for each experiment."

rsmtool/rsmsummarize.py
```diff
@@ -75,7 +77,15 @@ def check_experiment_dir(experiment_dir, configpath):
...
+        
+        return list(zip(jsons, [experiment_name]*len(jsons)))
```
desilinguist: At this point, `jsons` _has to_ be a list of one element right? Couldn't add a comment to that effect and rewrite this as:

```suggestion
        return [jsons[0], experiment_name]
```

Or am I missing something? I guess I don't know what happens in the case where `experiment_name` evaluates to `False` and there are multiple JSONs in the experiment directory.

rsmtool/rsmsummarize.py
```diff
@@ -129,15 +139,29 @@ def run_summary(config_file_or_obj, output_dir):
         else:
             configpath = os.getcwd()
 
+    logger.info('Saving configuration file.')
+    configuration.save(output_dir)
```
desilinguist: Same question as before. Were we just not doing this before?

rsmtool/rsmsummarize.py
```diff
@@ -129,15 +139,29 @@ def run_summary(config_file_or_obj, output_dir):
...
+
     # get the list of the experiment dirs
     experiment_dirs = configuration['experiment_dirs']
 
+    # Get experiment names if any
+    experiment_names = configuration.get('experiment_names')
+    if experiment_names:
+        dirs_with_names = zip(experiment_dirs, experiment_names)
+    else:
+        dirs_with_names = zip(experiment_dirs, [None]*len(experiment_dirs))
```
desilinguist: Just a thought, it might be a bit more readable to rewrite this block as:

```python
experiment_names = configuration.get('experiment_names')
experiment_names = experiment_names if experiment_names else [None]*len(experiment_dirs)
dirs_with_names = zip(experiment_dirs, experiment_names)
```

tests/test_configuration_parser.py
```diff
@@ -252,10 +252,33 @@ def test_validate_config_experiment_id_9(self):
...
+                'experiment_dirs': ["dir1", "dir2", "dir3"],
+                'experiment_names': ['exp1', 'exp2', 'exp3', 'exp4']}
+
+        # Add data to `ConfigurationParser` object
+        self.parser.load_config_from_dict(data)
+        self.parser.validate_config(context='rsmsummarize')
+
+
+    @raises(ValueError)
+    def test_validate_config_too_many_experiment_names(self):
```
desilinguist: This has a duplicated name. Should this be `test_validate_config_too_few_experiment_names`?

tests/test_experiment_rsmsummarize.py
```diff
@@ -119,3 +120,18 @@ def test_run_experiment_summary_no_json():
                        source,
                        'rsmsummarize.json')
     do_run_summary(source, config_file)
+
+
+@raises(FileNotFoundError)
```
desilinguist: Why does this test raise a `FileNotfoundError` if the test name ends with "_too_many_jsons()`?

rsmtool/rsmsummarize.py
aloukina: If you have a single experiment in a folder and did not ask for an experiment name (default behavior), this would give you `[(exp_id, None)]`. If you have multiple experiments in the same folder (undocumented option), experiment name will have to be `None` and this will give you `[(exp1, None), (exp2, None), (exp3, None)]`. 

I added a comment to explain this.

rsmtool/rsmsummarize.py
aloukina: No :)

rsmtool/rsmcompare.py
aloukina: We were not doing this before: see https://github.com/EducationalTestingService/rsmtool/issues/194

tests/test_experiment_rsmsummarize.py
aloukina: Should be ValueError

rsmtool/configuration_parser.py
aloukina: I decided that `validate_config` was the best place to make sure the number of experiment names matches the number of experiments since this is where we check for other abnormalities in the config file.  To do this, I needed to first make sure both fields are correctly parsed as lists (`process_config`) which is why I switched the order of the two operations.

In fact, some of our existing checks in `validate_config` also implicitly assumed the correct typing of the fields. E.g. check #9 which looks at whether `predict_expected_scores` is set to `True` only for models that support it. Previously, if you supplied `{'model': 'NNLR', 'predict_expected_scores': 'false'}`,  `false` would be treated as  a string, `if new_config[predict_expected_scores]` would be interpreted as `True` and the check would fail since `NNLR` does not support `predict_expected_scores`.

This is a very edge case but it illustrates that the new order is better aligned with our logic. I added a few tests for this too.

Issue #262, opened 2019-06-17
Make tolerance for trimming a config setting. 


aloukina:

aloukina:
Addressed by #264

PR #264, opened 2019-06-21
Allow custom trim tolerance

aloukina:
This PR addresses #262 : allow users set tolerance for trimming as a setting in the `trim_tolerance` field in the config-file (`rsmtool` and `rsmeval` only). We also keep track of `trim_tolerance` in the `post-processing` parameters. 

* Updated code and notebooks
* Updated documentation
* Updated tests

While this is a minor functionality change, it affects several functions such as filter metrics so @jbiggsets needs to review to make sure it doesn't break things on your end.


rsmtool/configuration_parser.py
```diff
@@ -446,7 +446,28 @@ def check_flag_column(self,
...
+
+        Returns
+        -------
+        spec_trim_min : float
+            Specified trim min value
+        spec_trim_max : float
+            Specified trim max value
+        """
+        logging.warning("This function has been replaced by  "
+                        "get_trim_min_max_tolerance() and will "
```
jbiggsets: This is nit-picky, but it should be "This _method_ has been replaced ...".

rsmtool/preprocessor.py
```diff
@@ -2463,13 +2475,31 @@ def process_data_rsmpredict(self, config_obj, data_container_obj):
         h1_mean = df_postproc_params['h1_mean'].values[0]
         h1_sd = df_postproc_params['h1_sd'].values[0]
 
+        #if the we are using a newly trained model, use trim_tolerance from the
+        # df_postproc_params. If not, set it to the default value and show
+        # deprecation warning
+        if 'trim_tolerance' in df_postproc_params.columns:
```
jbiggsets: Also nit-picky, but this can just be ` if 'trim_tolerance' in df_postproc_params`

rsmtool/preprocessor.py
```diff
@@ -2463,13 +2475,31 @@ def process_data_rsmpredict(self, config_obj, data_container_obj):
...
+                            "compatibility.")
+
+
         # now generate the predictions for the features using this model
         logged_str = 'Generating predictions'
         logged_str += ' (expected scores).' if predict_expected_scores else '.'
         logging.info(logged_str)
+
+        # compute minimum and maximum score for expected predictions
+        min_score = int(np.rint(trim_min - trim_tolerance))
```
jbiggsets: If you're doing `np.rint()`, do you need to cast these to `int()` afterward?

rsmtool/modeler.py
```diff
@@ -1305,18 +1309,21 @@ def predict_train_and_test(self,
                                                                        train_predictions_sd,
                                                                        human_labels_mean,
                                                                        human_labels_sd,
-                                                                       trim_min, trim_max)
+                                                                       trim_min, trim_max,
```
desilinguist: Separate lines?

rsmtool/modeler.py
```diff
@@ -1305,18 +1309,21 @@ def predict_train_and_test(self,
...
+                                                                       trim_tolerance)
 
         logging.info('Processing test set predictions.')
         df_test_predictions = FeaturePreprocessor.process_predictions(df_test_predictions,
                                                                       train_predictions_mean,
                                                                       train_predictions_sd,
                                                                       human_labels_mean,
                                                                       human_labels_sd,
-                                                                      trim_min, trim_max)
+                                                                      trim_min, trim_max,
```
desilinguist: Here too?

rsmtool/preprocessor.py
aloukina: Surprisingly, yes: 

```In [2]: a = np.rint(5.4)

In [3]: a
Out[3]: 5.0

In [4]: type(a)
Out[4]: numpy.float64

In [5]: type(int(a))
Out[5]: int```

PR #265, opened 2019-07-30
Minor edit to documentation

aloukina:
I found an error in documentation about transformations and corrected it. I also added further detail.


doc/config_rsmtool.rst
```diff
@@ -200,7 +200,9 @@ If this option is set to ``true`` the system will try apply feature transformati
     * ``addOneInv``: 1/(x+1)
     * ``addOneLn``: ln(x+1)
 
-Note that ``inv`` is never used for features with positive values. Defaults to ``false``.
+Only transformations that produce numeric results for *all* values for a given feature will be considered. Thus for example if a feature has a single negative value, `sqrt` transformation will be ignored even if it would have resulted in the highest correlation with human score for the remaining values.  In addition,  ``inv`` and ``addOneInv`` transformations are never used for features with both positive and negative values.
```
desilinguist:
```suggestion
We only consider transformations that produce numeric results for *all* values for a given feature column. For example, if a feature column contains a single negative value, the ``sqrt`` transformation will be ignored even if it would have resulted in the highest correlation with human score for the remaining values.  In addition, the ``inv`` and ``addOneInv`` transformations are never used for feature columns that contain both positive and negative values.
```

PR #268, opened 2019-08-06
Updating partial correlations function

jbiggsets:
This PR fixes a relatively minor bug in the `partial_correlations()` function. 

Essentially, when the determinant of the covariance matrix is very close to (but not exactly) zero, `np.linalg.inv()` will still calculate an answer, but the matrix it returns will be ill-formatted. To fix this, we should instead use the pseudo-inverse in cases where the determinant is less than the lowest-representable 32 bit integer.

This is how the `pcor` function in R works. See the issue here for further details:

https://github.com/EducationalTestingService/factor_analyzer/issues/38


rsmtool/utils.py
```diff
@@ -470,19 +470,27 @@ def partial_correlations(df):
...
+            # set the entire matrix to NaNs
+            if np.linalg.det(df_cov) < np.finfo(np.float32).eps:
+                icvx = np.linalg.pinv(df_cov)
+                warnings.warn('The inverse of the variance-covariance matrix '
+                              'was calculated using the Moore-Penrose generalized '
+                              'matrix inversion, due to its determinant being at '
+                              'or very close to zero.')
+            else:
+                icvx = np.linalg.inv(df_cov)
+        except:
```
desilinguist: I am not a fan of this particular way of doing this especially since using `except` without a specific exception isn't a good idea. Can't we structure this as a `try ... except` with an `AssertionError` for the determinant and then whatever specific exception you were trying to catch after that?

rsmtool/utils.py
```diff
@@ -429,7 +429,7 @@ def covariance_to_correlation(m):
     if not numrows == numcols:
         raise ValueError('Input matrix must be square')
 
-    Is = np.sqrt(np.abs(1 / np.diag(m)))
+    Is = np.sqrt(1 / np.diag(m))
```
desilinguist: Why are we removing `np.abs` here?

jbiggsets: We shouldn't need `np.abs()` anymore, and it shouldn't exist (if we're mirroring the `poor()` function). 

The reason we were getting negative numbers here was because some of the covariance matrices that we were inverting had determinants  _extremely_ small (and sometimes negative), but not quite zero. So, we weren't getting a `LinAlgError` errors from `inv()`, but we were getting ill-conditioned matrices. 

Using `pinv()` in these cases should resolve the problem, and we shouldn't get negative elements in the diagonal even after inversion.

rsmtool/utils.py
jbiggsets: Sure, I can restructure the code this way. I'll also add some more tests.

PR #270, opened 2019-08-07
Bug in rsmsummarize when trim min is null

aloukina:
The latest changes to add `trim_tolerance` introduced a bug in RSMSummarize: if the user has RSMTool deduct `trim min` and `trim max` from the data, these would be set to `null` in the config file and this would cause an error in one of the RSMSummarize notebooks. I have now fixed it and added a test.


rsmtool/notebooks/summary/evaluation.ipynb
```diff
@@ -39,11 +39,33 @@
     "            df_eval['system score type'] = 'scale' if config.get('use_scaled_predictions') == True or config.get('scale_with') is not None else 'raw'        \n",
     "            \n",
     "            # get truncation ranges\n",
-    "            trim_min = float(config.get('trim_min'))\n",
-    "            trim_max = float(config.get('trim_max'))\n",
-    "            trim_tolerance = config.get('trim_tolerance', 0.49998)\n",
-    "            min_score = trim_min - trim_tolerance\n",
-    "            max_score = trim_max + trim_tolerance\n",
+    "            # first let's see if we have them in the config file\n",
```
desilinguist:
```suggestion
    "            # first let's assume that we have them either in the config file or the params file\n",
```

rsmtool/notebooks/summary/evaluation.ipynb
```diff
@@ -39,11 +39,33 @@
...
+    "            \n",
+    "            try: \n",
+    "                trim_min = float(config.get('trim_min'))\n",
+    "                trim_max = float(config.get('trim_max'))\n",
+    "                trim_min_known = True\n",
```
desilinguist: I think we can just move this statement to before the `try` block ...

rsmtool/notebooks/summary/evaluation.ipynb
```diff
@@ -39,11 +39,33 @@
...
+    "            except TypeError: \n",
+    "                # if these are Nones, get them from postprocessing params\n",
+    "                postproc_params_file = os.path.join(csvdir, '{}_postprocessing_params.{}'.format(model_id,\n",
+    "                                                                                                 file_format))\n",
+    "                if os.path.exists(postproc_params_file):\n",
+    "                    df_postproc_params = DataReader.read_from_file(postproc_params_file)\n",
+    "                    trim_min = df_postproc_params['trim_min'].values[0]\n",
+    "                    trim_max = df_postproc_params['trim_max'].values[0]\n",
+    "                    trim_min_known = True\n",
```
desilinguist: ... and get rid of this one.

jbiggsets: What about something like this instead?

```
    "            # truncation ranges\n",
    "            # get the trim tolerance, min, and max from the config\n",
    "            trim_tolerance = config.get('trim_tolerance', 0.49998)\n",
    "            trim_min = config.get('trim_min')\n",
    "            trim_max = config.get('trim_max')\n",
    "            # if the trim min and max are not None, then use them to calculate min and max scores\n",
    "            if trim_min is not None and trim_max is not None:\n",
    "                min_score = trim_min - trim_tolerance\n",
    "                max_score = trim_max + trim_tolerance\n",
    "            # otherwise, use the postprocessing parameters file, if it exists, or default min and max to '?'\n",
    "            else:\n",
    "                postproc_file = os.path.join(csvdir, '{}_postprocessing_params.{}'.format(model_id, file_format))\n",
    "                if os.path.exists(postproc_file):\n",
    "                    df_postproc = DataReader.read_from_file(postproc_file)\n",
    "                    min_score = df_postproc['trim_min'].values[0] - trim_tolerance\n",
    "                    max_score = df_postproc['trim_max'].values[0] + trim_tolerance\n",
    "                else:\n",
    "                    min_score, max_score = '?', '?'\n",
```

desilinguist: Don't you need to convert `trim_min` and `trim_max` to floats first?

jbiggsets: I'll add that, just to be safe. (Although, I thought these would always be numeric or null by the time they were saved to the JSON.)

rsmtool/notebooks/summary/evaluation.ipynb
```diff
@@ -38,13 +38,25 @@
...
+    "                max_score = float(trim_max) + trim_tolerance\n",
+    "            # otherwise, use the postprocessing parameters file, if it exists, or default min and max to '?'\n",
+    "            else:\n",
+    "                postproc_file = os.path.join(csvdir, '{}_postprocessing_params.{}'.format(model_id, file_format))\n",
+    "                if os.path.exists(postproc_file):\n",
+    "                    df_postproc = DataReader.read_from_file(postproc_file)\n",
+    "                    min_score = df_postproc['trim_min'].values[0] - trim_tolerance\n",
+    "                    max_score = df_postproc['trim_max'].values[0] + trim_tolerance\n",
+    "                else:\n",
+    "                    min_score, max_score = '?', '?'\n",
```
desilinguist: Isn't this something that should _never_ happen or do we have a way to handle these later in the code?

rsmtool/notebooks/summary/evaluation.ipynb
```diff
@@ -37,44 +40,53 @@
...
-    "    return(df_evals)\n",
+    "    return df_evals, has_missing_trims\n",
+    "\n",
+    "df_eval, has_missing_trims = read_evals(model_list, file_format_summarize)\n",
     "\n",
-    "df_eval = read_evals(model_list, file_format_summarize)\n",
+    "if has_missing_trims:\n",
+    "    display(Markdown('**Note:** Minimum and maximum truncation scores could not '\n",
+    "                     'be located in some cases. When truncation values are missing, '\n",
+    "                     'they will be set to `?`.'))\n",
```
desilinguist: 1. I think we need to say _why_ they could not be located. 
2. Are the terms "Minimum and maximum truncation scores" and "truncation values" referring to the same entities? If so, perhaps a better name for said entities would be something like "The minimum and maximum scores after truncation"?

rsmtool/notebooks/summary/evaluation.ipynb
```diff
@@ -37,29 +40,54 @@
...
     "    else:\n",
     "        df_evals = pd.DataFrame()\n",
-    "    return(df_evals)\n",
+    "    return df_evals, has_missing_trims\n",
     "\n",
-    "df_eval = read_evals(model_list, file_format_summarize)\n",
+    "df_eval, has_missing_trims = read_evals(model_list, file_format_summarize)\n",
+    "\n",
+    "if has_missing_trims:\n",
+    "    display(Markdown('**Note:** The minimum and maximum scores after truncation could not be '\n",
```
desilinguist:
```suggestion
    "    display(Markdown('**Note:** The minimum and/or maximum scores after truncation could not be '\n",
```

rsmtool/notebooks/summary/evaluation.ipynb
```diff
@@ -37,29 +40,54 @@
...
+    "                     'be located in some cases. This is because the `trim_min` or `trim_max` '\n",
```
desilinguist:
```suggestion
    "                     'be computed in some cases. This is because `trim_min` and/or `trim_max` '\n",
```

rsmtool/notebooks/summary/evaluation.ipynb
```diff
@@ -37,29 +40,54 @@
...
+    "                     'could not be found in either the configuration file or the postprocessing '\n",
+    "                     'parameters file. Scores that could not be located will be set to `?`.'))\n",
```
desilinguist:
```suggestion
    "                     'parameters file. Scores that could not be computed are shown as`?`.'))\n",
```

jbiggsets:
Just a note that this PR morphed a bit:
- Updated the logic for extracting trim values in the summary evaluation notebook.
- Added an additional test to ensure that this new logic was working correctly.
- Added a message when some min and max values cannot be calculated.
- Fixed an existing bug in two tests where the `Configuration` object was being passed.

Issue #244, opened 2018-11-05
support .jsonlines input format


aoifecahill:
It would be nice if rsmtool could support the .jsonlines input format that skll also supports. This would make it easier to get some of the additional benefits that rsmtool offers over skll without having to convert data files (sometimes from sparse to dense formats).

aloukina:
@aoifecahill We wouldn't be able to support sparse formats properly since any response with null values for any feature is removed from training/evaluation. Do you ever use jsonlines for non-sparse data?

aoifecahill:
Hmm, we do actually use jsonlines for non-sparse data sometimes.

aloukina:
@aoifecahill Can you think of a situation where you might want output files as `.jsonlines` ?

aoifecahill:
Not off the top of my head. We are used to getting the outputs in .csv and so our scripts are set up for that.

desilinguist:
Addressed by #271

PR #271, opened 2019-08-08
Allow jsonlines as input

aloukina:
Addresses #244 
* jsonlines supported as input format but not as output format as there doesn't seem to be a use case for that. 
- Updated code
- Added tests
- Updated documentation

* Also updated code in reporter so that fairness notebook is only included when we have subgroups.


rsmtool/reader.py
```diff
@@ -22,6 +22,47 @@
...
+         Data frame containing the data in the given file.
+    """
+
+    df = pd.read_json(filename,
+                      orient='records',
+                      lines=True,
+                      dtype=converters)
+        
+    # let's check if we have nested columns
+    nested_cols = [c for c in df.columns if isinstance(df[c][0],dict)]
```
jbiggsets: What if the first row in the column is null?

rsmtool/reader.py
```diff
@@ -22,6 +22,47 @@
...
+
+    if len(nested_cols) > 0:
+        dfs = []
+        for col in nested_cols:
+            df_nested_column = pd.io.json.json_normalize(df[col])
+            dfs.append(df_nested_column)
+            df.drop(col, axis=1, inplace=True)
+        df_new = pd.concat([df] + dfs, axis=1)
```
jbiggsets: If we just assign this back to `df`, can we get rid of the `else` below?

doc/pipeline.rst
```diff
@@ -22,14 +22,21 @@ In addition to the HTML report, RSMTool also saves the intermediate outputs of a
...
+The input files containing feature values and scores for all responses in training and evaluation data should be either in tabular format with features and scores stored in columns and each row correponding to a single response or in ``jsonlines`` format with one line per response. See below for further description of ``.jsonlines`` format. 
 
 RSMTool supports input files in ``.csv``, ``.tsv``, ``.sas7bdat``, ``xls``/``.xlsx``, or ``.jsonlines`` format. The format of the file is determined based on the extension. In all cases the :ref:`output files<intermediate_files_rsmtool>` will be saved in ``.csv`` format by default (see :ref:`file format <file_format>` for other intermediate file output options).
 
-For *Excel* spreadsheets all data must be stored in the first sheet. 
+For ``Excel`` spreadsheets all data must be stored in the first sheet. 
 
-Jsonlines: 
+Each line in ``.jsonlines`` format should correspond to one response represented as a dictionary with column names as keys and values as values. For example:
```
desilinguist:
```suggestion
In a ``.jsonlines`` format file, each line corresponds to a response and is represented as a dictionary with column names as the keys and column values as the values. For example:
```

doc/pipeline.rst
```diff
@@ -22,14 +22,21 @@ In addition to the HTML report, RSMTool also saves the intermediate outputs of a
...
 
+.. literalinclude:: ../tests/data/files/train.jsonlines
+   :lines: 1,2
+
+RSMTool allows nested ``jsons`` in each line of ``.jsonlines`` file with no more than one level of nesting as in the following example. Note that only the lower level keys will be used when parsing the files and processing the data. Thus in the example below second level labels ``x`` and ``metadata`` will be ignored.
```
desilinguist:
```suggestion
Although RSMTool does allow for a single level of nesting in the JSON objects on each line of a ``.jsonlines`` format file, *only* the top-level keys are actually used when parsing the files and processing the data. Therefore, in the example bellow, the keys ``x`` and ``metadata`` will be ignored. 
```

rsmtool/reader.py
```diff
@@ -22,6 +22,47 @@
 from rsmtool.container import DataContainer
 
 
+
+def read_jsonlines(filename, converters=None):
+    """ Read jsonlines from a file.
```
desilinguist:
```suggestion
    """
    Read jsonlines from a file. 
```

rsmtool/reader.py
```diff
@@ -22,6 +22,47 @@
...
+    Normalize nested jsons with up to one level of nesting
+
+    Parameters
+    ----------
+    filename: str
+        Name of file to read
+    converters : dict or None, optional
+        A dictionary specifying how the types of the columns
+        in the file should be converted. Specified in the same
+        format as for `pd.read_csv()`.
```
desilinguist: We need a link to `read_csv()` in the pandas documentation.

rsmtool/reader.py
```diff
@@ -22,6 +22,47 @@
...
+            df_nested_column = pd.io.json.json_normalize(df[col])
+            dfs.append(df_nested_column)
+            df.drop(col, axis=1, inplace=True)
+        df_new = pd.concat([df] + dfs, axis=1)
+    else:
+        df_new = df
+        
+    return df_new
+
+
```
desilinguist: Here's a more readable (and probably a bit more efficient) way of doing this I think. It also addresses @jbiggsets's concerns I think. This is supposed to replace the bits after the `pd.read_json()` call:

```python
dfs = []   
for column in df:
    try:
        df_column = pd.io.json.json_normalize(df[column])
    except AttributeError:
        df_column = df[column].copy()

    dfs.append(df_column)

df = pd.concat(dfs, axis=1)
```

rsmtool/reader.py
aloukina: Can you give an example?

rsmtool/reader.py
aloukina: Not sure - I think different var names make the code more readable but it's probably a judgement call. @desilinguist ?

doc/pipeline.rst
```diff
@@ -22,14 +22,23 @@ In addition to the HTML report, RSMTool also saves the intermediate outputs of a
...
-For *Excel* spreadsheets all data must be stored in the first sheet. 
+For ``Excel`` spreadsheets, all data must be stored in the first sheet. 
 
-Jsonlines: 
+In a ``.jsonlines`` format file, each line corresponds to a response and is represented as a dictionary with column names as the keys and column values as the values. For example:
 
+.. literalinclude:: ../tests/data/files/train.jsonlines
+   :lines: 1,2
+
+Although RSMTool does allow for nesting in the JSON objects on each line of a ``.jsonlines`` format file, the top-level keys will be ignored when parsing the files and processing the data. Therefore, in the example bellow, the keys ``x`` and ``metadata`` will be ignored.
```
desilinguist:
```suggestion
Although RSMTool does allow for nesting in the JSON objects on each line of a ``.jsonlines`` format file, the top-level keys will be ignored when parsing the files and processing the data. Therefore, in the example below, the keys ``x`` and ``metadata`` will be ignored. 
```

doc/pipeline.rst
```diff
@@ -22,14 +22,23 @@ In addition to the HTML report, RSMTool also saves the intermediate outputs of a
...
 
+.. literalinclude:: ../tests/data/files/train.jsonlines
+   :lines: 1,2
+
+Although RSMTool does allow for nesting in the JSON objects on each line of a ``.jsonlines`` format file, the top-level keys will be ignored when parsing the files and processing the data. Therefore, in the example bellow, the keys ``x`` and ``metadata`` will be ignored. 
+
+.. literalinclude:: ../tests/data/files/train_nested.jsonlines
+   :lines: 1,2
+
+If the data contains nested records with more than two levels, the column names for nested records beyond the top level will be generated using ``.`` to separate levels: ``{'foo': {'bar': {'foo2': 0, 'foo3': 0}}`` -> ``foo`` will be ignored, the columns will be named ``bar.foo2`` and ``bar.foo3``.
```
desilinguist:
```suggestion
If the file contains nesting of more than two levels, the column names for nested records beyond the top level will be generated using ``.`` to separate levels: For example, given the JSON object ``{'foo': {'bar': {'foo2': 0, 'foo3': 0}}``,  ``foo`` will be ignored and the columns will be named ``bar.foo2`` and ``bar.foo3``.
```

rsmtool/reader.py
```diff
@@ -11,17 +11,64 @@
...
+    """
+
+    df = pd.read_json(filename,
+                      orient='records',
+                      lines=True,
+                      dtype=converters)
+
+    # make sure we didn't get a plain json
+    if type(df.columns) == pd.RangeIndex:
+        raise ValueError("It looks like {} is a simple json file. "
```
desilinguist: Hmm, how do we distinguish between a "simple json" file and a jsonlines file with a single instance?

rsmtool/reader.py
desilinguist: I think you need the link here too?

tests/test_reader.py
```diff
@@ -261,3 +276,99 @@ def test_setup_none_in_path(self):
...
+                                {'values': {'id': '002',
+                                           'feature1': 2,
+                                           'feature2': 2.5}},
+                                {'values': {'id': '003',
+                                           'feature1': 3,
+                                           'feature2': 3.5}}]
+        self.check_jsonlines_output(all_nested_jsonlines)
+
+    def test_read_jsonlines_more_than_2_levels(self):
+        mlt_nested_jsonlines = [{'values': {'id': '001',
```
desilinguist:
```suggestion
        multi_nested_jsonlines = [{'values': {'id': '001',
```

Just a nitpick about variable naming.

tests/test_reader.py
```diff
@@ -261,3 +276,99 @@ def test_setup_none_in_path(self):
...
+                                {'values': {'id': '002',
+                                           'features': {'feature1': 2,
+                                                        'feature2': 2.5}}},
+                                {'values': {'id': '003',
+                                           'features': {'feature1': 3,
+                                                        'feature2': 3.5}}}]
+        self.expected.columns = ['id',
+                                 'features.feature1',
+                                 'features.feature2']
+        self.check_jsonlines_output(mlt_nested_jsonlines)
```
desilinguist:
```suggestion
        self.check_jsonlines_output(multi_nested_jsonlines)
```

rsmtool/reader.py
desilinguist: Got it. Thanks!

rsmtool/reader.py
```diff
@@ -11,17 +11,72 @@
...
+        df = pd.read_json(filename,
+                          orient='records',
+                          lines=True,
+                          dtype=converters)
+    except ValueError:
+        raise ValueError("The jsonlines file is not formatted correctly. "
+                         "Please check that each line ends with a comma, "
+                         "there is no comma at the end of the last line, "
+                         "and that all quotes match. Please also check that"
+                         "Nones are written as `Null` and not `NaN.")
```
desilinguist:
```suggestion
                         "any undefined values are written out as `null` and not `NaN`.")
```

Issue #299, opened 2019-09-30
Degrees of freedom in QWK computation


aloukina:
* Switch QWkappa computation to ddof=0. We confirmed that this way the outputs of the two formulas will match and the output for discreet values will match that by Sklearn or skll.
* Mention ddof explicitly in documentation
* Explain the differences between sklearn and skll implementation.

aloukina:
Addressed by #300

PR #300, opened 2019-09-30
Updated ddof in kappa computation.

aloukina:
This addresses #299. 
I updated the formula, added wording to documentation, updated test data and also added new tests to check that the output for discrete values matches that from skll/sklearn.


rsmtool/utils.py
```diff
@@ -762,8 +762,9 @@ def quadratic_weighted_kappa(y_true_observed, y_pred, ddof=1):
     ddof : int, optional
         Means Delta Degrees of Freedom. The divisor used in
         calculations is N - ddof, where N represents the
-        number of elements.
-        Defaults to 1.
+        number of elements. When ddof is 0 to zero, the results
```
desilinguist:
```suggestion
        number of elements. When ddof is set to zero, the results
```

doc/evaluation.rst
```diff
@@ -91,19 +91,26 @@ where:
...
 
 
-Unlike :ref:`Cohen's kappa<kappa>` which is only computed for rounded scores, quadratic weighted kappa is computed for continous scores using the following formula: 
+Unlike :ref:`Cohen's kappa<kappa>` which is only computed for rounded scores, quadratic weighted kappa is computed for continuous scores using the following formula: 
 
 
 :math:`QWK=\frac{E[M-H]^2}{Var(H)+Var(M)+(\bar{M}-\bar{H})^2}`
 
-QWK is computed using :ref:`rsmtool.utils.quadratic_weighted_kappa<qwk_api>` with ``ddof`` set to ``1``.
+Note that in this case the variance is the population variance and thus it is computed by dividing by ``N`` and not ``N-1`` as in other cases.
```
desilinguist:
```suggestion
Note that in this case the variance is the population variance and, thus, it is computed by dividing by ``N`` and not by ``N-1``, as in other cases.  
```

doc/evaluation.rst
```diff
@@ -91,19 +91,26 @@ where:
 
 Kappa is computed using `skll.metrics.kappa <https://skll.readthedocs.io/en/latest/api/skll.html#from-metrics-module>`_ with ``weights`` set to ``None`` and ``allow_off_by_one`` set to ``False`` (default).
 
+.. note::
+    See `this discussion <https://github.com/EducationalTestingService/skll/issues/391#issuecomment-444145567>`_ for the explanation of how `skll <https://skll.readthedocs.io/en/latest/api/skll.html#from-metrics-module>`_ implementation differs from the `sklearn implementation <https://scikit-learn.org/stable/modules/generated/sklearn.metrics.cohen_kappa_score.html>`_. The two implementations might produce different results if the matrix contains missing labels, for example ``1, 2, 4``. In ``skll`` implementation the missing ``2``  will be automatically added to the list of labels. ``sklearn`` implementation would only add ``2`` if full label list is passed to the function as the optional ``labels`` parameter.
```
desilinguist:
```suggestion
   See `this discussion <https://github.com/EducationalTestingService/skll/issues/391#issuecomment-444145567>`_ for the explanation of how the `SKLL implementation <https://skll.readthedocs.io/en/latest/api/skll.html#skll.kappa>`_  differs from the `scikit-learn implementation <https://scikit-learn.org/stable/modules/generated/sklearn.metrics.cohen_kappa_score.html>`_. The two implementations might produce different results if the matrix contains missing labels. For example, consider the hypothetical scenario where our predictions only contain the labels ``1``, ``2``, and ``4``. In the SKLL implementation, the missing ``3``  will be automatically added to the list of labels whereas in the scikit-learn implementation, the ``3`` would only be added if a complete list of labels was passed to the function via the optional ``labels`` keyword argument.
```

Issue #297, opened 2019-09-24
Generate requirements in meta.yaml from setup.py?


jbiggsets:
Right now, we're explicitly listing out all of the requirements in the `meta.yaml` file:

https://github.com/EducationalTestingService/rsmtool/blob/f5b63fa4324a7c4cdf4e382f085348943cec15ad/conda-recipe/rsmtool/meta.yaml#L22-L46

Some of these (like `sphinx`) we have to list out explicitly, but the rest we could probably populate directly from `setup.py` file, so we don't have to update both of these things and make sure they're consistent with each release.

I believe we can use this approach instead:

```
{% set data = load_setup_py_data() %}

...

{% block requirements -%}
requirements:
  build:
    {% for req in data.get('install_requires', []) -%}
    - {{(req.replace(">", " >") if ">" in req else req.replace("==", " "))}}
    {% endfor %}
  run:
    - python
    {% for req in data.get('install_requires', []) -%}
    - {{(req.replace(">", " >") if ">" in req else req.replace("==", " "))}}
    {% endfor %}
{%- endblock %}
```

Thoughts?

https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#id15

desilinguist:
Ooh, this looks like a great idea! Have you already tested it and it works?

jbiggsets:
Testing it out now. 👍

desilinguist:
Addressed by #305

PR #305, opened 2019-10-09
Update conda recipe and requirements

jbiggsets:
This PR is intended to address #297, but it includes several additional changes that should to be reviewed closely.

Here is a quick rundown of the changes:

1. **Removed `conda_requirements.txt`**. Previously, we were maintaining two sets of requirements: `requirements.txt` and `conda_requirements.txt`. The second was being used to create `conda` environments (using `conda create -n foo --file conda_requirements.txt`). However, since these sets of requirements were virtually identical and both can be used to create `conda` environments, I discarded the `conda_requirements.txt` file.  The one major difference between these files is that `conda_requirements.txt` had `zeromq`. This cannot be in the `requirements.txt` file, since it leads to an error when using `pip install -e .`. I think this is all right, since it's not entirely necessary (and may be installed by `Jupyter` anyway?), but we can add it to the `meta.yaml` separately, if it's a problem. (See below on the updates to `meta.yaml`.)

2. **Switched to `requirements.txt` in `.travis.yml`**. Related to the above, our `travis.yml` file had previously used `conda_requirements.txt` to set up the `conda` environment for our builds. This has been changed to use `requirements.txt`  in our Travis build.

3. **Updated `requirements.txt`**. Some minor tweaks were also made to the `requirements.txt` file, mostly reordering the list of requirements. I also removed `scikit-learn` (since this is already a `skll` dependency).

4. **Updated `meta.yaml` to use `setup.py`**. Previously, the `meta.yaml` file was listing all of the requirements manually, which meant we had to ensure these requirements were always up-to-date. I changed this configuration file to instead rely on the `setup.py` data, and pull the requirements directly from `install_requires`. This update takes advantage of [`Jinja` templating](https://conda.io/projects/conda-build/en/latest/resources/define-metadata.html#templating-with-jinja). I tested this on my Mac, and it worked. I am going to test it on a Windows machine today, and it would probably be good to have others test it on their machines, as well.

5. **Re-generated `environment.yml`**. Finally, I re-generated the `environment.yml` to ensure that it was consistent with our `requirements.txt` file.

6. **Add missing test file**. Added `test_container.py` to `.travis.yml` and Azure, per #310.


conda-recipe/rsmtool/meta.yaml
```diff
@@ -19,55 +21,19 @@ build:
...
-    - sphinx_rtd_theme
-    - zeromq
-    - coverage
-    - parameterized
-    - setuptools
-    - openpyxl
-    - xlrd
-    - xlwt
-
+    - python 3.6.8
```
desilinguist: I think it might be nice to build `noarch` packages that can still pin Python versions. For example, see [here](https://github.com/JRCSTU/jrcstu-forge/commit/12b0695738ece2e5d03e16e319de81a4be62ead5). Also, I don't think we want to pin to a specific version of Python 3.6. Any Python >= 3.6 should work fine, including Python 3.7 right?

conda-recipe/rsmtool/meta.yaml
```diff
@@ -19,55 +21,19 @@ build:
...
-    - statsmodels>=0.9.0,<0.9.1
-    - sphinx
-    - sphinx_rtd_theme
-    - zeromq
-    - coverage
-    - parameterized
-    - openpyxl
-    - xlrd
-    - xlwt
+    - python 3.6.8
```
desilinguist: Same here.

environment.yml
```diff
@@ -3,26 +3,24 @@ channels:
...
-- parameterized
-- zeromq
-- sphinx
-- sphinx_rtd_theme
-- coverage
-- setuptools
-- openpyxl
-- xlrd
-- xlwt
+  - python=3.6.8
```
desilinguist: Again, let's not be too specific here.

environment.yml
```diff
@@ -3,26 +3,24 @@ channels:
...
+  - ipython=6.5.0
+  - jupyter=1.0.0
+  - joblib=0.11
+  - matplotlib=2.1.2
+  - nose=1.3.7
+  - notebook=5.7.2
+  - numpy=1.14.0
```
desilinguist: Do we need to pin numpy? Since this is only used for readthedocs, using the latest version of numpy should be fine right? And same for `pandas` etc.?

requirements.txt
```diff
@@ -4,18 +4,17 @@ joblib==0.11
 matplotlib==2.1.2
 nose==1.3.7
 notebook==5.7.2
-numpy==1.14.*
+numpy==1.14.0
```
desilinguist: Why this change?

requirements.txt
```diff
@@ -4,18 +4,17 @@ joblib==0.11
...
 pandas==0.23.4
```
desilinguist: Do things break if we use the latest version of `pandas`?

conda-recipe/rsmtool/meta.yaml
jbiggsets: I agree with everything, except I wasn't sure about 3.7. I've never tested anything with 3.7, and the old `conda_requirements.txt` had `python>=3.6.3,<3.7`.

requirements.txt
jbiggsets: Not sure. I'll test it out. :)

requirements.txt
jbiggsets: I don't have strong feelings about this, but I'm always paranoid about differences in even minor version of `numpy` leading to small differences in results.

desilinguist:
@jbiggsets what's the status of this PR?

requirements.txt
jbiggsets: Yes, they do.

conda-recipe/rsmtool/meta.yaml
```diff
@@ -9,7 +11,7 @@ build:
   number: 0
   script:
     - cd $SRC_DIR
-    - $PYTHON setup.py install
+    - $PYTHON setup.py install --single-version-externally-managed --record=record.txt
```
desilinguist: I think you don't need to do this since its use is discouraged. See the SKLL recipe for a much easier alternative.

desilinguist:
We should also check that we aren't bundling test data (which tends to be large) into our conda or PyPI package.

jbiggsets:
> We should also check that we aren't bundling test data (which tends to be large) into our conda or PyPI package.

Why would that be the case, and how would we check that?

desilinguist:
> > We should also check that we aren't bundling test data (which tends to be large) into our conda or PyPI package.
> 
> Why would that be the case, and how would we check that?

In SKLL, when I just did `python setup.py install`, it included tests, test data, examples, and example data in the conda package. I think we prevented it in SKLL, through a combination of the `MANIFEST.in` file and through appropriate fields in `setup.py`.

desilinguist:
The conda package was tested and worked fine on both Linux and Windows as seen [here](https://github.com/EducationalTestingService/rsmtool-conda-tester/pull/5). Therefore, I am now merging this PR.

PR #322, opened 2019-12-05
Better handling of long feature names

aloukina:
This PR addresses #64 : I added an extra check for long feature names and different layout for some of the plots in case at least one feature names is longer than 10 characters. There is a also a new test with long feature names.


rsmtool/notebooks/header.ipynb
```diff
@@ -350,6 +352,8 @@
     "    if exists(feature_file):\n",
     "        df_features = DataReader.read_from_file(feature_file, converters=converter_dict)\n",
     "        features_used = [c for c in df_features.feature.values]\n",
+    "        # compute the longest feature name: we'll need if for the plots\n",
+    "        longest_feature_name = max(map(len, features_used))\n",
```
jbiggsets: What if there's no `feature_file`? Doesn't that mean the `longest_feature_name` variable won't exist?

aloukina: That's a good question: we have the check because the header is shared across tools ('contexts') and for example feature_file is not present in RSMEval outputs. 

As things stand now, the feature file will always be present in an RSMTool experiment which is the only context where we would also include the feature analysis notebooks. The definition of `features_used` variable in l.352 follows the same logic.

jbiggsets: Got it. Thanks! :)

rsmtool/notebooks/dff_by_group.ipynb
```diff
@@ -45,26 +45,47 @@
...
     "    if len(df_train_preproc_selected) > 0:\n",
+    "        \n",
+    "        # we need to reduce col_wrap and increase width if the feature names are too long\n",
+    "        if longest_feature_name > 10:\n",
+    "            col_wrap = 2\n",
+    "            # adjust height to allow for wrapping really long names. We allow 0.25 in per line\n",
+    "            height = 2+(math.ceil(longest_feature_name/30)*0.25)\n",
+    "            aspect = 5/height\n",
+    "            # show legend near the second plot\n",
+    "            plot_with_legend = 1\n",
```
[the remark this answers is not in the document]
aloukina: The index of the plot next to which we will add a legend

rsmtool/notebooks/feature_descriptives.ipynb
```diff
@@ -53,8 +53,16 @@
...
-    "aspect = 3 if len(features_used) > 40 else 2\n",
+    "\n",
+    "# we need to increase the plot height if feature names are long\n",
+    "if longest_feature_name > 10:\n",
+    "    height = 3 + math.ceil((longest_feature_name - 10)/10)\n",
+    "else:\n",
+    "    height = 3\n",
+    "    \n",
+    "# we also need a higher aspect if we have more than 40 features\n",
+    "aspect = 9/height if len(features_used) > 40 else 6/height\n",
```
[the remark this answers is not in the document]
aloukina: We can but they are more or less a guesstimate of what looks good

Issue #355, opened 2020-01-31
Allow run_* functions to take dictionary as input


aloukina:
Currently the main API functions `run_experiment`, `run_evaluation` etc. take either file path to the config file or a valid `Configuration` object with all missing fields set to default values. Constructing and validating such object requires an in-depth understanding of the interplay between `Configuration` and `ConfigurationParser`.

It will be simpler to allow user pass a file path, object or a dictionary that will then be normalized, vadlidated and converted to a Configuration object.

desilinguist:
Addressed by #356 and #371.

PR #356, opened 2020-01-31
Better input handling in RSMTool

aloukina:
1. This PR introduces more consistent treatment of how we store configuration path by removing `.filepath` attribute and replacing it with `.configdir` and `.filename`. It fixes the bug in #225 where RSMTool crashed if Configuration object was passed without `.filepath` attribute. I also added test for this. 

2. The users can now pass dictionary as input to `run_experiment` script (#355 ).

3. I did some minor refactoring on the test code to avoid duplication and added tests for new functions in `test_utils`. 

I have only implemented the changes for RSMTool. The changes for other tool will be very similar so I would like to iron out any major issues we might find.  Please review them now and I will then implement the final version for other tools either in this or different PR.


rsmtool/rsmtool.py
```diff
@@ -65,24 +72,34 @@ def run_experiment(config_file_or_obj,
...
+
+        # initialize the parser from dict
+        parser = ConfigurationParser()
+        parser.load_config_from_dict(config_file_or_obj_or_dict)
+        configuration = parser.normalize_validate_and_process_config()
+        # set configpath to current directory
+        configpath = getcwd()
+        configuration.filepath = configpath
 
     else:
```
mulhod: Is there a reason not to _also_ check `isinstance(config_file_or_obj_or_dict, Configuration)`? In some cases where I've had this sort of `if`/`elif`/`else`, I like to make sure about what I'm interacting with (or try to at least) in each piece and then leave the `else` part to deal with an "unrecognized" value-type error and raise an exception.

rsmtool/test_utils.py
```diff
@@ -71,21 +73,29 @@ def check_run_experiment(source,
         Path where the test experiments are located. Unless specified, the
         rsmtool test directory is used. This can be useful when using these
         experiments to run tests for RSMExtra.
+    input_is_file: bool, optional
+        Input to the `run_experiment` script should be a f file path
```
mulhod: typo: remove "f"

rsmtool/test_utils.py
```diff
@@ -40,7 +40,9 @@ def check_run_experiment(source,
                          consistency=False,
                          skll=False,
                          file_format='csv',
-                         given_test_dir=None):
+                         given_test_dir=None,
+                         input_is_file=True,
```
mulhod: It seems like these two parameters (`input_is_file` and `config_obj_or_dict`) could be collapsed in the some way. Maybe getting rid of `input_is_file` and retaining only the `config_obj_or_dict` and then using its value:

```python
    if config_obj_or_dict is None:
        config_input = join(test_dir,
                           'data',
                           'experiments',
                           source,
                           '{}.json'.format(experiment_id))
    else:
        config_input = config_obj_or_dict
```

rsmtool/rsmtool.py
aloukina: Good point. Also added a test for this.

rsmtool/configuration_parser.py
```diff
@@ -689,6 +888,54 @@ def load_config_from_dict(self, config_dict):
                                  'been assigned. You cannot assign another'
                                  'dictionary.')
 
+        if configdir is None:
+            configdir = getcwd()
+
+        self._configdir = abspath(configdir)
+        logging.info("Reference directory configdir was set to {}".format(self._configdir))
```
desilinguist:
```suggestion
        logging.info("Configuration directory will be set to {}".format(self._configdir))
```

let's be consistent with the other logging message.

rsmtool/configuration_parser.py
```diff
@@ -689,6 +888,54 @@ def load_config_from_dict(self, config_dict):
...
+        Load configuration dictionary.
+
+        Parameters
+        ----------
+        config_dict : dict
+            A dictionary containing the configuration
+            parameters to parse.
+        configdir : str, optional
+            Path to the reference directory used to resolve
+            any relative path in the dictionary. By default
```
desilinguist:
```suggestion
            any relative path in the dictionary. 
```

rsmtool/configuration_parser.py
```diff
@@ -689,6 +888,54 @@ def load_config_from_dict(self, config_dict):
...
+            will be set to current workind directory
```
desilinguist:
```suggestion
            Defaults to the current working directory.
```

rsmtool/modeler.py
```diff
@@ -1342,7 +1342,9 @@ def predict_train_and_test(self,
         config_as_dict = configuration.to_dict()
         config_as_dict.update(new_config_dict)
 
-        configuration = Configuration(config_as_dict, configuration.filepath)
+        configuration = Configuration(config_as_dict,
+                                      configdir=configuration.configdir,
+                                      filename=configuration._filename)
```
desilinguist: This is a little weird to me. If we are going to access the private `_filename` attribute in such a public way then may be it should be a property too.

rsmtool/configuration_parser.py
```diff
@@ -44,33 +47,112 @@
...
+                                     "for filename.")
+                if 'configdir' in kwargs:
+                    raise ValueError("Configuration received both "
+                                     "positional argument for filepath "
+                                     "and keyword argument for configdir")
+                # raise deprecation warning
+                warnings.warn("Starting from RSMTool 8.0, you will need to "
+                              "specify configdir as a keyword-only argument "
+                              "when initializing a Configuration object ",
+                              DeprecationWarning)
```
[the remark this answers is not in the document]
aloukina: It will raise a warning, if you pass a second positional argument to `Configuration` (current API): e.g. `Configuration(dict, 'path_to_file'). It will raise a warning if you pass a second positional argument plus any of the new keyword arguments. In other words, if you combine old and new API.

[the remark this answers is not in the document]
aloukina: You can have a look at the tests too.

rsmtool/configuration_parser.py
```diff
@@ -155,27 +237,81 @@ def __iter__(self):
...
         new_path : str
             A new file path for the
             configuration object.
         """
-        self._filepath = new_path
+        warnings.warn("The `filepath` attribute of the Configuration "
+                      "object will be removed n RSMTool 8.0 "
+                      "use `configdir` and `_filename` if you "
+                      "need to set a new path to the "
+                      "configuration file", DeprecationWarning)
```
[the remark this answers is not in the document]
aloukina: [note to self: to address after the first round of commits]

rsmtool/configuration_parser.py
```diff
@@ -602,6 +739,60 @@ def _check_config_is_loaded(self):
...
+        warnings.warn("The `_filepath` attribute of the "
+                      "ConfigurationParser object "
+                      "ill be removed n RSMTool 8.0 "
+                      "use `_configdir` and `_filename` if you "
+                      "need to set a new path to the "
+                      "configuration file", DeprecationWarning)
+        new_filename = basename(new_path)
+        new_configdir = dirname(abspath(new_path))
+        self._filename = new_filename
+        self._configdir = new_configdir
```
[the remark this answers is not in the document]
aloukina: I agree in general. I added this in case internal ETS tools set or retrieve `._filepath` to avoid breaking things without warning.

rsmtool/modeler.py
aloukina: Agree, let me make it a property. 
This line is necessary to make sure it's not lost when we re-instantiate the object.

rsmtool/configuration_parser.py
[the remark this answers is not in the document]
desilinguist: I don't think anyone is using this, TBH. And, even if they were, they shouldn't have been so it'd be good to get them to fix it.

rsmtool/test_utils.py
```diff
@@ -782,6 +790,55 @@ def check_subgroup_outputs(output_dir, experiment_id, subgroups, file_format='cs
...
+def copy_data_files(temp_dir_name,
+                   input_file_dict,
+                   given_test_dir=None):
+    """
+    Copy input files from tests/data into
+    a temporary directory specified by the user.
+    This is a utility function for testing
+    situations where RSMTool is expected to use
+    current directory as a reference directory
+    for resolving paths in the configuration
```
desilinguist: How about?

```
A utility function to copy files from the ``tests/data`` directory into 
a specified temporary directory. Useful for tests where the 
current directory is to be used as the reference for resolving paths
in the configuration.
```

rsmtool/test_utils.py
```diff
@@ -782,6 +790,55 @@ def check_subgroup_outputs(output_dir, experiment_id, subgroups, file_format='cs
             ok_(length == partition_info.iloc[0][group])
 
 
+def copy_data_files(temp_dir_name,
+                   input_file_dict,
```
desilinguist:
```suggestion
                    input_file_dict,
```

rsmtool/test_utils.py
```diff
@@ -782,6 +790,55 @@ def check_subgroup_outputs(output_dir, experiment_id, subgroups, file_format='cs
...
+                   given_test_dir=None):
```
desilinguist:
```suggestion
                    given_test_dir=None):
```

rsmtool/test_utils.py
```diff
@@ -782,6 +790,55 @@ def check_subgroup_outputs(output_dir, experiment_id, subgroups, file_format='cs
...
+    situations where RSMTool is expected to use
+    current directory as a reference directory
+    for resolving paths in the configuration
+
+    Parameters
+    ----------
+    temp_dir_name : str
+        Name of the temporary directory.
+    input_file_dict : dict
+        A dictionary of files list of files to copy.
```
desilinguist:
```suggestion
        A dictionary of files to copy with keys as the file type
        and the values are their paths relative to the `tests`
        directory.
```

rsmtool/test_utils.py
```diff
@@ -782,6 +790,55 @@ def check_subgroup_outputs(output_dir, experiment_id, subgroups, file_format='cs
...
+        The keys show the type of file. The values should be
```
desilinguist:
```suggestion
```

rsmtool/test_utils.py
```diff
@@ -782,6 +790,55 @@ def check_subgroup_outputs(output_dir, experiment_id, subgroups, file_format='cs
...
+        paths to the files relative to `tests` directory.
```
desilinguist:
```suggestion
```

rsmtool/test_utils.py
```diff
@@ -782,6 +790,55 @@ def check_subgroup_outputs(output_dir, experiment_id, subgroups, file_format='cs
...
+    given_test_dir : str, optional
+        Path where the test experiments are located. Unless specified, the
```
desilinguist:
```suggestion
        Directory where the the test experiments are located. Unless specified, the
```

tests/test_configuration_parser.py
```diff
@@ -6,7 +6,11 @@
 import pandas as pd
 
 from io import StringIO
-from os.path import dirname, join
+from os import getcwd
+from os.path import (abspath,
+                     dirname,
+                     join)
```
desilinguist: I don't think you need newlines here.

tests/test_configuration_parser.py
```diff
@@ -92,6 +96,66 @@ def test_normalize_config(self):
             newdata = self.parser.normalize_config()
             ok_('use_scaled_predictions' not in newdata.keys())
 
+    def test_load_config_from_dict(self):
+        data = {'expID': 'test'}
+        configdir = abspath('/path/to/configdir')
```
desilinguist: Isn't this already an absolute path?

tests/test_configuration_parser.py
```diff
@@ -507,6 +571,67 @@ def test_get_correct_configparser_json(self):
...
+    def test_init_with_configdir_only_as_kword_argument(self):
+        configdir = 'some/path'
+        config_dict = {'exp_id': 'my_experiment'}
+        config = Configuration(config_dict,
+                               configdir=configdir)
+        eq_(config._filename, None)
+        eq_(config._configdir, abspath(configdir))
+
+    @ raises(ValueError)
+    def test_init_with_filepath_as_positional_and_keyword(self):
```
desilinguist:
```suggestion
    def test_init_with_filepath_positional_and_filename_keyword(self):
```

tests/test_configuration_parser.py
```diff
@@ -507,6 +571,67 @@ def test_get_correct_configparser_json(self):
...
+        filepath = 'some/path/file.json'
+        config_dict = {'exp_id': 'my_experiment'}
+        _ = Configuration(config_dict,
+                          filepath,
+                          filename=filepath)
+
+    @ raises(ValueError)
+    def test_init_with_filepath_as_positional_and_configdir_keyword(self):
```
desilinguist:
```suggestion
    def test_init_with_filepath_positional_and_configdir_keyword(self):
```

tests/test_configuration_parser.py
```diff
@@ -694,17 +819,70 @@ def test_str_correct(self):
...
         eq_(config.__str__(), 'flag_column')
 
+    # this is a test for an attribute that will be
+    # deprecated
     def test_get_filepath(self):
-        filepath = '/path/to/file.json'
-        config = Configuration({"flag_column": "[advisories]"}, filepath)
-        eq_(config.filepath, filepath)
-
+        with warnings.catch_warnings(record=True) as w:
```
desilinguist:
```suggestion
        with warnings.catch_warnings(record=True) as warning_list:
```

tests/test_configuration_parser.py
```diff
@@ -694,17 +819,70 @@ def test_str_correct(self):
...
+            filepath = '/path/to/file.json'
+            config = Configuration({"flag_column": "[advisories]"},
+                                   configdir='/path/to/',
+                                   filename='file.json')
+            eq_(config.filepath, abspath(filepath))
+            assert len(w) == 1  # we get one deprecation warnings here
```
desilinguist:
```suggestion
            assert len(warning_list) == 1  # we get one deprecation warning here
```

tests/test_configuration_parser.py
```diff
@@ -694,17 +819,70 @@ def test_str_correct(self):
...
+            assert issubclass(w[-1].category, DeprecationWarning)
```
desilinguist:
```suggestion
            assert issubclass(warning_list[0].category, DeprecationWarning)
```

tests/test_configuration_parser.py
```diff
@@ -694,17 +819,70 @@ def test_str_correct(self):
...
+
+    # this is a test for an attribute that will be
+    # deprecated
     def test_set_filepath(self):
-        filepath = '/path/to/file.json'
-        new_file_path = 'path/that/is/new.json'
-        config = Configuration({"flag_column": "[advisories]"}, filepath)
-        config.filepath = new_file_path
-        eq_(config.filepath, new_file_path)
+        with warnings.catch_warnings(record=True) as w:
```
desilinguist:
```suggestion
        with warnings.catch_warnings(record=True) as warning_list:
```

tests/test_configuration_parser.py
```diff
@@ -694,17 +819,70 @@ def test_str_correct(self):
...
+            new_file_path = '/newpath/to/new.json'
+            config = Configuration({"flag_column": "[advisories]"},
+                                   configdir='/path/to/',
+                                   filename='file.json',)
+            config.filepath = new_file_path  # first deprecation warning
+            eq_(config._filename, 'new.json')
+            eq_(config.configdir, abspath('/newpath/to'))
+            eq_(config.filepath, abspath(new_file_path))  # second deprecation warning
+            assert len(w) == 2  # we get two deprecation warnings here
```
desilinguist:
```suggestion
            assert len(warning_list) == 2  # we get two deprecation warnings here
```

tests/test_configuration_parser.py
```diff
@@ -694,17 +819,70 @@ def test_str_correct(self):
...
+            assert issubclass(w[-2].category, DeprecationWarning)
```
desilinguist:
```suggestion
            assert issubclass(warning_list[0].category, DeprecationWarning)
```

tests/test_configuration_parser.py
```diff
@@ -694,17 +819,70 @@ def test_str_correct(self):
...
+            assert issubclass(w[-1].category, DeprecationWarning)
```
desilinguist:
```suggestion
            assert issubclass(warning_list[1].category, DeprecationWarning)
```

rsmtool/test_utils.py
```diff
@@ -782,6 +790,55 @@ def check_subgroup_outputs(output_dir, experiment_id, subgroups, file_format='cs
             ok_(length == partition_info.iloc[0][group])
 
 
+def copy_data_files(temp_dir_name,
```
desilinguist:
```suggestion
def copy_test_data_files(temp_dir_name,
```

tests/test_experiment_rsmtool_3.py
```diff
@@ -22,6 +27,7 @@
 else:
     from rsmtool.test_utils import rsmtool_test_dir
 
+DIRS_TO_REMOVE = []
```
desilinguist: What do you think about using [`tempfile.TemporaryDirectory()`](https://docs.python.org/3.8/library/tempfile.html#tempfile.TemporaryDirectory) that can be explicitly cleaned up after you are done using the `.cleanup()` method? IMO, much better than having global variables and special functions.

tests/test_configuration_parser.py
aloukina: Windows resolves them with `D:` so without this tests fail on Azure.

tests/test_test_utils.py
```diff
@@ -0,0 +1,29 @@
...
+class TestCopyData():
+
+    def setUp(self):
+        self.dirs_to_remove = []
+
+    def tearDown(self):
+        for d in self.dirs_to_remove:
+            shutil.rmtree(d)
+
+    def test_copy_test_data_files(self):
```
desilinguist:
```suggestion
    def test_copy_data_files(self):
```

tests/test_test_utils.py
```diff
@@ -0,0 +1,29 @@
...
+        file_dict = {'train': 'data/files/train.csv',
+                     'features': 'data/experiments/lr/features.csv'}
+        expected_dict = {'train': 'temp_test_copy_data_file/train.csv',
+                         'features': 'temp_test_copy_data_file/features.csv'}
+        self.dirs_to_remove.append('temp_test_copy_data_file')
+        output_dict = copy_test_data_files('temp_test_copy_data_file',
```
desilinguist:
```suggestion
        output_dict = copy_data_files('temp_test_copy_data_file',
```

Issue #351, opened 2020-01-30
LinearSVR fails to converge in new version


aloukina:
After switching to scipy 1.4.1 we get the following warning in tests:
`envs/rsmtool_skll/lib/python3.7/site-packages/sklearn/svm/base.py:929: ConvergenceWarning: Liblinear failed to converge, increase the number of iterations.`

I have re-run the tests using old environment and can confirm that this warning is new.

aloukina:
@desilinguist Any thoughts on this? We are using default settings when calling skll:

https://github.com/EducationalTestingService/rsmtool/blob/07f91264795dc8fb5de9518cb65ff4e1685836ef/rsmtool/modeler.py#L1057

desilinguist:
Yeah, this is coming from scikit-learn, I think. I thought we had addressed this in SKLL v2.0 but may be we missed it?

aloukina:
Do you have a test there that can check for this?

desilinguist:
We have tests for SVMs but I don't think it gets triggered for the test data there. Which RSMTool test is this?

aloukina:
linearsvr: https://github.com/EducationalTestingService/rsmtool/tree/master/tests/data/experiments/linearsvr

desilinguist:
I don't think there's anything that can be done for this other than just suppressing the warnings in the test for now. For a small dataset like the one we use in this test (N=500), when we do grid search, the model fit just doesn't converge in the default 1000 iterations and with the given tolerance. To get the warnings to suppress, we'll need to bump up the tolerance and increase the number of iterations (see [here](https://github.com/EducationalTestingService/rsmtool/files/4188178/linearsvr-rsmtool.pdf)).

In the future, when we can pass in SKLL parameters to RSMTool, we can set the hyper-parameters of the model to suppress the warning.

desilinguist:
Addressed by #361.

Issue #352, opened 2020-01-30
Add warning capture to tests


aloukina:
We generally suppress warnings for tests where these are expected but there are few exceptions. We want to deal with these so that by default we expect no warnings when running tests:

test_utils.test_standardized_mean_difference_zero_denominator 
test_utils.test_difference_of_standardized_means_with_no_population_info 
test_utils.test_partial_correlations_with_singular_matrix

aloukina:
I addressed some of these, but we need to fix #367 before this.

desilinguist:
Addressed by #369.

PR #361, opened 2020-02-12
Suppress convergence warnings for LinearSVR test (Also make future warning suppression easier)

desilinguist:
As discussed in #351, the current solution is to simply suppress the convergence warnings coming from scikit-learn until we can pass in custom values for SKLL hyperparameters.

In addition, I also made it much simpler to add future suppression warnings for tests based on `run_experiment`. I modified `check_run_experiment()` and `do_run_experiment()`  to accept a list of warning types as a keyword argument. This will allow any types of warnings to be suppressed and doesn't require adding individual `filterwarnings()` in each test.

### DIFF
--- a/rsmtool/test_utils.py
+++ b/rsmtool/test_utils.py
@@ -40,7 +40,8 @@ def check_run_experiment(source,
                          consistency=False,
                          skll=False,
                          file_format='csv',
-                         given_test_dir=None):
+                         given_test_dir=None,
+                         suppress_warnings_for=[]):
     """
     Function to run for a parameterized rsmtool experiment test.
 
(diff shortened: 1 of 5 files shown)


rsmtool/test_utils.py
```diff
@@ -40,7 +40,8 @@ def check_run_experiment(source,
                          consistency=False,
                          skll=False,
                          file_format='csv',
-                         given_test_dir=None):
+                         given_test_dir=None,
+                         suppress_warnings_for=[]):
```
aloukina: Nice! This will be really useful for https://github.com/EducationalTestingService/rsmtool/issues/352

aloukina:
File an issue for warning suppression to be consistent?

PR #372, opened 2020-02-20
Overhaul RTD building process

desilinguist:
- Rename `readthedocs.yml` to `.readthedocs.yml` for consistency with `.travis.yml` and `.pep8speaks.yml`.
- Switch to `pip` for building instead of `conda`.
- Move requirements for documentation building `doc/requirements.txt`.
- Use already installed `numpy`, `scipy`, and `pandas` packages in RTD image instead of installing those ourselves.
- Remove `environment.yml` file which was used only for conda.

This branch builds [successfully](https://rsmtool.readthedocs.io/en/fix-broken-rtd-builds/) on RTD.


.readthedocs.yml
```diff
@@ -0,0 +1,20 @@
...
+  configuration: doc/conf.py
+
+# Optionally build your docs in additional formats such as PDF and ePub
+formats: all
+
+# Optionally set the version of Python and requirements required to build your docs
+python:
+  version: 3.6
+  install:
+    - requirements: doc/requirements.txt
```
mulhod: Maybe I'm misunderstanding this, but how come these requirements don't include things like `sphinx` and `sphinx_rtd_theme`. Should I be able to create a Conda environment from this file alone and build the documentation?

mulhod: While on this point, why are those requirements included in the main `requirements.txt` anyway?

desilinguist: No, `doc/requirements.txt` is a requirements file that is specific for ReadTheDocs build which uses a docker image that already includes sphinx, numpy, scipy etc which is why we don't need them. 

To do local development (whether code or documentation), you need to use the main `requirements.txt` file which installs everything you need. 

Good question. I will add this to the contributing section.

Issue #357, opened 2020-02-07
Create a version of the documentation for the Dash app


desilinguist:
For those of us who use [Dash](https://kapeli.com/dash) on macOS, it would be neat to have a version of the RSMTool documentation available there. It would make it available offline, only a keyboard shortcut away, and even easier to navigate.

desilinguist:
Relevant links:

- https://github.com/Kapeli/Dash-User-Contributions#contribute-a-new-docset
- https://kapeli.com/docsets#improveDocset

desilinguist:
I have this pretty much working. PR coming soon.

PR #375, opened 2020-02-20
Add tooling to create Dash docset for RSMTool and other minor changes.

desilinguist:
This PR adds the tooling to create an RSMTool docset for the [Dash](https://kapeli.com/dash) app on macOS as per #357. Dash is an offline documentation browser that integrates with basically every IDE/editor out there and has docsets for pretty much any major open-source project under the sun. I use it heavily during development as do many other macOS folks. In addition to the Dash stuff, it also makes some other minor changes (listed below).

This PR: 
1. Adds the underlying machinery to make generating the Dash docset as easy as running `make dash` in the `doc` directory.
2. Adds instructions on how to then make that docset make available to the general public.
3. Adds notes to the repo README, the contributing guidelines, as well as the main page of the documentation to publicize that we (will) have such a docset available.
4. Cleans up unused options in `doc/conf.py`.
5. Fixes some issues with links, indentations, and newlines that sphinx was complaining about. 
6. Moves the static assets (images and the paper PDF) under `assets` and updates the documentation and `doc/conf.py`.
7. Modifies `doc/conf.py` to automatically add the current year as opposed to us needing to manually update it every year.
8. Excludes the recently added `deprecated_positional_argument()` decorator from `sphinx-autodoc`.

Notes:
-  Dash docset uses a different theme (Alabaster) since the RTD theme uses its own sidebar which looks weird since Dash has its own sidebar.
- Dash is a paid app but there's an open-source alternative called [Zeal](https://zealdocs.org) that works with Dash docsets. 

To see what this docset looks like in Dash, take a look at this [video](https://drive.google.com/open?id=12jKY3PDmcMHdk-OW6oIga4YlDusnF3Vw).


aloukina:
Thank you for the video!

doc/add_dash_anchors.py
```diff
@@ -0,0 +1,87 @@
...
+import logging
+import re
+import pathlib
+import unicodedata
+from urllib.parse import quote
+
+from bs4 import BeautifulSoup
+
+# pre-define the list of HTML files we want to miodify
+FILES_TO_MODIFY = ["advanced_usage.html", "api.html", "contributing.html",
```
aloukina: How do we make sure this list is up-to-date? Can we create it automatically based on directory content?

desilinguist: Yeah, I tried this but struggled with it. There are some auto-generated HTML files that we do not want to add Dash nchors to (for example, `genindex.html`). I haven’t been able to figure out if this set is fixed and so we can just include all files _except_ the ones in this set. I can work on this in the future.

desilinguist:
> I am slightly concerned that it adds extra overhead to the release process but we can over time whether we feel that the benefits overweigh that.

That’s fair enough. I’ll try to figure out how to automate this as much as possible as we go forward.

PR #378, opened 2020-02-23
Make tests run with package and add missing test file

desilinguist:
[This line](https://github.com/EducationalTestingService/rsmtool/blob/master/rsmtool/test_utils.py#L34) that computes `rsmtool_test_dir` basically assumes that the tests are always being run from the RSMTool source code directory. However, this is not true when we are testing the conda/pip packages (which do not include the tests). We currently use the `TESTDIR` environment variable to get around this but this does not always work as this PR shows (many of the tests still rely on `rsmtool_test_dir` directly). However, for now, this PR provides a fix for such tests.

This PR:

1. Add the missing test file `test_test_utils.py` to Travis and Azure builds.
2. Modify `copy_data_files()` and some other test functions to explicitly specify `rsmtool_test_dir`so that the tests pass when running the tests with packages.
3. Includes `rsmtool/test_utils.py` for coverage checking – it was being excluded until now.
4. A few other fixes (e.g., Windows path issues, extra slashes etc.) to get the tests to pass.

I discovered these issues when testing packages through the pip and conda testing builds which were failing. With these changes, the builds now pass as seen below (I modified those builds to check out this branch temporarily instead of `master`):

[pip package on windows](https://dev.azure.com/EducationalTestingService/web/build.aspx?pcguid=f4f8d738-e70a-421a-9cc1-983ff2abe8d4&builduri=vstfs%3a%2f%2f%2fBuild%2fBuild%2f335)
[pip package on linux](https://travis-ci.org/EducationalTestingService/rsmtool-pip-tester/builds/654077330)
[conda package on windows](https://dev.azure.com/EducationalTestingService/web/build.aspx?pcguid=f4f8d738-e70a-421a-9cc1-983ff2abe8d4&builduri=vstfs%3a%2f%2f%2fBuild%2fBuild%2f336)
[conda package on linux](https://travis-ci.org/EducationalTestingService/rsmtool-conda-tester/builds/654077419)

Once this PR is merged, I will re-run the package builds and then submit the release branch PR.

**Note**: the decrease in coverage is expected since we’re now including `test_utils.py`.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -8,4 +8,3 @@ omit =
     */site-packages/nose/*
     */unittest2/*
     *rsmtool/notebooks/templates/report.tpl
-    *rsmtool/test_utils.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ env:
   matrix:
   - TESTFILES="tests/test_experiment_rsmtool_1.py"
   - TESTFILES="tests/test_comparer.py tests/test_configuration_parser.py tests/test_experiment_rsmtool_2.py"
-  - TESTFILES="tests/test_analyzer.py tests/test_experiment_rsmeval.py tests/test_fairness_utils.py tests/test_prmse_utils.py tests/test_container.py test_test_utils.py"
+  - TESTFILES="tests/test_analyzer.py tests/test_experiment_rsmeval.py tests/test_fairness_utils.py tests/test_prmse_utils.py tests/test_container.py tests/test_test_utils.py"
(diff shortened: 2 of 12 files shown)


desilinguist:
Here's how we test the packages: we create a conda environment, install the built package via conda or pip, check out the RSMTool code and then the tests that are checked out there. More details [here](https://github.com/educationalTestingService/rsmtool-conda-tester), for example.

mulhod:
Thanks for the info! Didn't know this other repo existed!

Issue #384, opened 2020-02-26
Allow run_* functions to accept Path object


aloukina:
In the past the scripts would accept Path object (undocumented option). Now this no longer works.

PR #386, opened 2020-02-26
Allow run functions to accept pathlib.Path objects

aloukina:
This PR addresses #384. 
- I refactored the code since the input processing part was shared by all components: this is now done via Configuration Parser method. 
- The path to the config file can be passed as either string or pathlib.Path object. Previously the latter functionality was available as undocumented option. 
- I didn't add new tests for that method since the functionality is already tested in tests we wrote for individual components

### DIFF
--- a/rsmtool/configuration_parser.py
+++ b/rsmtool/configuration_parser.py
@@ -26,12 +26,14 @@ from os.path import (abspath,
                      join,
                      splitext)
 
+from pathlib import Path
 from ruamel import yaml
 
 from rsmtool import HAS_RSMEXTRA
 from rsmtool.utils import parse_json_with_comments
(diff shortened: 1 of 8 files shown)


rsmtool/configuration_parser.py
```diff
@@ -764,15 +766,15 @@ def _check_config_is_loaded(self):
 
 
     @classmethod
-    def get_configparser(cls, filepath, *args, **kwargs):
+    def get_configparser(cls, config_file_or_obj_or_dict, *args, **kwargs):
```
desilinguist: Hmm, might it make more sense to just move this logic to the `ConfigurationParser()` initializer instead of having this classmethod since all we are doing is getting the args and then calling the initializer anyway?

Issue #370, opened 2020-02-14
Replace all path operations and functions with `pathlib`


desilinguist:
Now that we are Python 3.6+ only, it makes more sense to use the more readable `pathlib.Path` interface rather than os-level functions.

aloukina:
Also related to #385

PR #388, opened 2020-03-03
Refactor `configuration_parser.py`

desilinguist:
**[Note: This PR supersedes #386].**

In our discussions, we recently found `configuration_parser.py` to be significantly over-engineered and a bit unwieldy to use. This PR attempts to fix that via a refactoring of the two main classes: `Configuration` and `ConfigurationParser`. There are also resulting changes to tests and other modules. Specifically:

1. The `Configuration` class is now instantiated using a dictionary so there's no need for special methods to deal with dictionaries. 

2. `ConfigurationParser` is now used only to parse configuration files (JSON or CFG). You create an instance for a given file and then call `.parse()` that returns a `Configuration` instance. The methods `normalize_config()`, `process_config()`, and `validate_config()` are all now class methods that accept a dictionary as input and return a dictionary as output. The private attribute `_config` no longer exists which makes operation explicit and not via side-effects. 

3. The classes `JSONConfigurationParser` and `CFGConfigurationParser` no longer exist. They have been replaced with private methods for `ConfigurationParser` called `_parse_json_file()` and `_parse_cfg_file()`. 

4. When `ConfigurationParser.parse()` is called, it calls the appropriate private method to read in the file into a Python dictionary. Then it normalizes, processes, and validates that dictionary. Finally, it instantiates a Configuration object using the validated dictionary and returns it.

5. If a `Configuration` object is instantiated directly using the initializer, the input dictionary is also normalized, processed, and validated. This makes it so that experiment configurations are _always_ normalized, processed, and validated whether used via a file or directly.

6. There's a new function called `configure()` added to `configuration_parser.py` that contains the shared code from all of the tools for dealing with different types of inputs (strings, paths, objects, and dictionaries). [This is the part from  #386 essentially].

7. There were a few places in `preprocessor.py` and `modeler.py` where we were creating new Configuration objects entirely. This led to errors in this new scenario where everything is validated because there are some internal configuration fields that do not need to be validated. To address this, we replaced this with the following: first call `configuration.copy()` (which creates a deep copy) and then add each of the internal options to the copy using `__setitem__()` which bypasses the validation. This is okay because this is _only_ for internal use.

8. There were a few tests where `trim_min` and `trim_max` are stored as strings and then being accessed as `config[trim_min]` and `config[trim_max]`. Previously, they were being converted to floats via side-effects of point 7 above. However, once we got rid of that entirely, tests started failing. The answer here is to _always_ use `Configuration.get_trim_min_max_tolerance()` method rather than trying to access the fields directly.  

9. The tests in `test_configuration_parser.py` needed to be refactored to make sure that all of the required fields are present since all the `Configuration` objects created directly are now validated.

10. If `Configuration()` was created from a dictionary and `filename` was not specified, trying to access the `filepath` attribute later errors out. I modified it so that this now explicitly raises an exception. I also added a new test for this.

11. Several other tests needed to be changed to use the new API where possible.

12. Remove some uneeded imports and tweak some docstrings here and there.

I don't believe any documentation changes are required (the API documentation will be handled by sphinx autodoc) but please let me know if I missed something.

### DIFF
--- a/rsmtool/configuration_parser.py
+++ b/rsmtool/configuration_parser.py
@@ -19,6 +19,7 @@ from copy import copy, deepcopy
 from collections import Counter
 from configparser import ConfigParser
 from os import getcwd
+from os.path import abspath
 
 from pathlib import Path
 from ruamel import yaml
@@ -386,7 +387,10 @@ class Configuration:
 
         if new_path is None:
             raise ValueError("The `configdir` attribute cannot be set to `None` ")
-        self._configdir = Path(new_path).resolve()
+
+        # TODO: replace `Path(abspath(new_path))` with `Path(new_path).resolve()
+        # once this Windows bug is fixed: https://bugs.python.org/issue38671
+        self._configdir = Path(abspath(new_path))
 
     @property
     def filename(self):
--- a/rsmtool/configuration_parser.py
+++ b/rsmtool/configuration_parser.py
@@ -18,12 +18,8 @@ import warnings
 from copy import copy, deepcopy
 from collections import Counter
 from configparser import ConfigParser
+from os import getcwd
 
-from os import getcwd, makedirs
-from os.path import (abspath,
-                     basename,
-                     dirname,
-                     join)
 from pathlib import Path
 from ruamel import yaml
 
@@ -42,7 +38,7 @@ from skll import Learner
 from skll.metrics import SCORERS
 
 if HAS_RSMEXTRA:
-    from rsmextra.settings import (default_feature_subset_file,
+    from rsmextra.settings import (default_feature_subset_file, # noqa
                                    default_feature_sign)
 
 
@@ -80,8 +76,8 @@ def deprecated_positional_argument():
                 # split filepath into
                 # configdir and filename
                 filepath = args[-1]
-                kwargs['filename'] = basename(filepath)
-                kwargs['configdir'] = dirname(abspath(filepath))
+                kwargs['filename'] = Path(filepath).name
+                kwargs['configdir'] = Path(filepath).resolve().parent
                 # remove filepath from positional arguments
                 args = args[:-1]
             return f(*args, **kwargs)
@@ -215,10 +211,10 @@ class Configuration:
 
         # set configdir to `cwd` if not given and let the user know
         if configdir is None:
-            configdir = getcwd()
+            configdir = Path(getcwd())
             logging.info("Configuration directory will be set to {}".format(configdir))
         else:
-            configdir = abspath(configdir)
+            configdir = Path(configdir).resolve()
 
         self._config = configdict
         self._configdir = configdir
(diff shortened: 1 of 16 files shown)


rsmtool/configuration_parser.py
```diff
@@ -897,41 +969,31 @@ def load_normalize_and_validate_config_from_dict(self,
...
-
-
-    def read_config_from_file(self, filepath):
-        """
-        Read the configuration file.
+        configuration : Configuration
+            A Configuration object containing the parameters in the
+            file that we instantiated the parser for.
+        """
+        _, extension = splitext(self._filename)
```
aloukina: Do we want to use Pathlib here since we are refactoring anyway?

rsmtool/configuration_parser.py
```diff
@@ -90,6 +90,77 @@ def wrapper(*args, **kwargs):
     return decorator
 
 
+def configure(tool_name, config_file_or_obj_or_dict):
```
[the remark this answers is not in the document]
desilinguist: Nothing except readability. Configuring a tool sounds better to me than Configuring a context but I am not married to it certainly :)

rsmtool/configuration_parser.py
```diff
@@ -249,13 +322,17 @@ def filepath(self):
...
             The path for the config file.
         """
-        warnings.warn("The `filepath` attribute of the Configuration "
-                      "object will be removed in RSMTool v8.0."
-                      "Use the `configdir` and `filename` attributes if you "
-                      "need the full path to the "
-                      "configuration file", DeprecationWarning)
-        filepath = join(self.configdir, self.filename)
-        return filepath
+        if not self.filename:
```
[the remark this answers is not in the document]
desilinguist: Yes, good catch.

rsmtool/configuration_parser.py
desilinguist: Note that `parse()` can only be used on an instantiated `ConfigurationParser` object and `__init__()` takes care of handling Paths. So, not sure the you would ever _not_ get a filename here.

aloukina: I meant: do we want to continue use `os` modules or switch to exclusively `pathlib` as per https://github.com/EducationalTestingService/rsmtool/issues/370 ?

desilinguist: 🤦‍♂. Yes, good idea :)

rsmtool/configuration_parser.py
```diff
@@ -973,7 +977,7 @@ def parse(self, context='rsmtool'):
             A Configuration object containing the parameters in the
             file that we instantiated the parser for.
         """
-        _, extension = splitext(self._filename)
+        extension = Path(self._filename).suffix.lower()
         filepath = join(self._configdir, self._filename)
```
aloukina:
```suggestion
        filepath = Path(self._configdir) / self._filename
```

aloukina: We should really just store `_configdir` as a `Path` object.

desilinguist: I like it! Let me do that.

desilinguist: Actually looks like `_configdir` is already a Path since it's currently defined as:

```
self._configdir = pathlike.resolve().parent
```

PR #395, opened 2020-03-05
Distribute `utils.py` functions into various `rsmtool.utils` package files.

desilinguist:
This PR resolves #394. 

- Create a `utils` directory and make it a package by adding a blank `__init__.py` to it.
- Move functions from `utils.py` to various files under the `utils` package.
- Tweak imports on all affected code, test and notebook files. Use relative imports in code files per Python best practices.
- Update API documentation to fix function paths and to add them to the API change note at the top.
- Add `# noqa`  comments so that linter doesn't complain in `__init__.py`.

Issue #354, opened 2020-01-30
Add config generator


aloukina:
Several uses expressed interest in config generator. The concern is that it is sometimes hard to remember all fields and there names. 

We could have a script that you run as

`generate_config rsmtool --groups --consistency >config.json` 

which generates a json file with all fields. The defaults would be prefilled and the obligatory fields will be filled with "ENTER_INFO_HERE". It would also be really helpful to list all sections since it will make it easy to delete the ones you don't want to run.

In addition to CL script, I'd love this functionality exposed via API for using in Python notebooks.

The `--groups` and `--consistency` flags are necessary since some of the fields are conditioned on whether you want these. 

Another idea we discussed is to have the script prompt for user input (e.g. to allow tab completion for paths).

desilinguist:
IMHO, it might be better to add a `quickstart` option to each of the tools rather than a separate script; something like:

```
rsmtool quickstart --with-groups --with-consistency > config.json
```

The implementation of this functionality could (and should) certainly be shared between the various utilities but I think in terms of exposing it to the user, this might be easier to remember. 

And if the user wants be prompted for each option, we could simply do:

```
rsmtool quickstart --interactive --with-groups --with-consistency
```

Thoughts?

desilinguist:
Another thought - if we are going to use a pre-filled dummy value, it might be useful to have the respective script be aware of the value so that instead of just failing, it could say something like "it seems you are using a config generated via quickstart but you forgot to fill out some fields".

aloukina:
Would `rsmtool quickstart ...` generate config and exit?

desilinguist:
Yes. Does "quickstart" indicate to you that it should also run the config?

aloukina:
To me it seems counterintuitive that the same command (`rsmtool`) is used for both running the experiment and generating config.

desilinguist:
Ah, I see your point and that does seem like a valid concern. There are two ways to handle this:

(1) We move to `rsmtool` being the driver script and have it take a "command" as to what it's supposed to do. For example:

```
$> rsmtool run <config.json>
$> rsmtool quickstart --with consistency --with groups > config.json
```

We'd obviously have a backwards-compatible path, e.g.,  "run" could be the default command and old invocations would work just fine. This is the approach followed in many other large Python CLI projects and would give us a good expansion path for future commands. 

(2) We stick with your suggestion and just tweak it a bit:

```
$> quickstart --for rsmtool --with sources --with consistency > config.json
[...]

PR #398, opened 2020-03-10
Enhanced command line interface for all tools and automated generation of tool configurations

desilinguist:
**Note**: This PR partially addresses #354 and should be reviewed after #395 is merged. The part of #354 that is left is the interactive mode which will be handled in a separate PR to make review easier.

In this PR:

1. I refactored the command line interfaces of all 5 tools by factoring out the command-line parser creation into a separate function `rsmtool.utils.commandline.setup_rsmcmd_parser()`. By doing so, we make it much easier to extend and modify the command-line interface to all utilities at the same time without needing to duplicate code. In addition, it also improves the consistency among the tools. I also added a named tuple called `CmdOption` to make it easy to specify non-standard options like in `rsmpredict`.

2. I added the ability to automatically generate configuration files in each tool by adding a subcommand called "generate". So, for example, to generate an rsmtool configuration file, you would simply run `rsmtool generate`. This outputs a well-indented JSON file to stdout containing useful comments and also outputs a warning to stderr saying  that the file must be edited and cannot be used as-is. This subcommand also supports the `--groups` flag which will add the subgroup-related sections to the `general_sections` lists; they are not otherwise added. All of this is made possibly by the `rsmtool.utils.commandline.generate_configuration()` function. Note the `suppress_warnings` keyword argument in that function. That argument is hooked up to a `--quiet` / `-q` switch for the `generate` subcommand for every tool. 

3. To make it consistent, the normal path of running an experiment for each command line tool is now part of the "run" subcommand. So, for example, to run an rsmtool experiment, you would run `rsmtool run config.json /path/to/dir`. However, I wanted to allow the old-style invocations to work without medication too so I made `run` the default sub-command so commands like `rsmtool config.json /path/to/dir` still work fine.

4. As part of the refactoring, I moved the check for whether the configuration file exists inside `ConfigurationParser.__init__()`. This really simplifies the code in the `main()` function of all tools since this bit is also shared among all of them.

5. I also moved the check for whether the output directory can be overwritten to the respective API functions as part of the refactoring. where appropriate. So, for example, `run_experiment()` now has an optional keyword argument called `overwrite_output` with a default value of `False`. This means that if you were expecting the API function to overwrite output in previous versions, you will now need to be explicit about it by passing `overwrite_output=True`.

6. To facilitate automated config generation, I modified `Configuration.__str__()` to output a full JSON dump instead of just the keys which makes a lot more sense. I also modified `Configuration.save()` to first call __str__ and then simply write the dump tp the file.

7. Add new test classes to `test_utils.py` for `CmdOption`, `generate_configuration()`, and `setup_rsmcmd_parser()`. Also added expected JSON outputs for `generate_configuration()`.

8. Tweak some of the logging setup in the 5 command line tools to get around the fact that in "run" mode, we usually log to stdout but in "generate" mode we do not want to log the warning to stdout since that would
make it part of the generated configuration which would make the JSON unparseable.

9. Added a brand new test file `test_cli.py` to test the command-line invocations for all 5 tools. Incorporated this new test file into the Travis and Azure CI builds.

10. Update API documentation to include `rsmtool.utils.commandline,.generate_configuration()`. Update documentation for all tools to include automated config generation. 

11. Minor cleanups here and there to docstrings, unneeded imports, and unneeded variables.


doc/config_rsmcompare.rst
```diff
@@ -3,7 +3,7 @@
 Experiment configuration file
 """""""""""""""""""""""""""""
 
-This is a file in ``.json`` format that provides overall configuration options for an ``rsmcompare`` experiment. Here's an example configuration file for `rsmcompare <https://github.com/EducationalTestingService/rsmtool/blob/master/examples/rsmcompare/config_rsmcompare.json>`_.
+This is a file in ``.json`` format that provides overall configuration options for an ``rsmcompare`` experiment. Here's an `example configuration file <https://github.com/EducationalTestingService/rsmtool/blob/master/examples/rsmcompare/config_rsmcompare.json>`_ for ``rsmcompare``. To make it easy to get started with  ``rsmcompare``, we provide a way to automatically generate a configuration file that you can then just edit based on your data and your needs. To do so, simply run ``rsmcompare generate`` at the commmand line. If you know you have :ref:`subgroups <subgroups_rsmtool>` in your data that you want to include in your analyses, run ``rsmcompare generate --groups`` instead. Next, we describe all of the ``rsmcompare`` configuration fields in detail.
```
aloukina:
```suggestion
This is a file in ``.json`` format that provides overall configuration options for an ``rsmcompare`` experiment. Here's an `example configuration file <https://github.com/EducationalTestingService/rsmtool/blob/master/examples/rsmcompare/config_rsmcompare.json>`_ for ``rsmcompare``. To make it easy to get started with  ``rsmcompare``, we provide a way to automatically generate a configuration file that you can then edit based on your data and your needs. To do so, simply run ``rsmcompare generate`` at the commmand line. If you have :ref:`subgroups <subgroups_rsmtool>` in your data that you want to include in your analyses, run ``rsmcompare generate --subgroups`` instead. Next, we describe all of the ``rsmcompare`` configuration fields in detail.
```

doc/config_rsmeval.rst
```diff
@@ -3,7 +3,7 @@
 Experiment configuration file
 """""""""""""""""""""""""""""
 
-This is a file in ``.json`` format that provides overall configuration options for an ``rsmeval`` experiment. Here's an example configuration file for `rsmeval <https://github.com/EducationalTestingService/rsmtool/blob/master/examples/rsmeval/config_rsmeval.json>`_.
+This is a file in ``.json`` format that provides overall configuration options for an ``rsmeval`` experiment. Here's an `example configuration file <https://github.com/EducationalTestingService/rsmtool/blob/master/examples/rsmeval/config_rsmeval.json>`_ for ``rsmeval``. To make it easy to get started with  ``rsmeval``, we provide a way to automatically generate a configuration file that you can then just edit based on your data and your needs. To do so, simply run ``rsmeval generate`` at the commmand line. If you know you have :ref:`subgroups <subgroups_eval>` in your data that you want to include in your analyses, run ``rsmeval generate --groups`` instead. Next, we describe all of the ``rsmeval`` configuration fields in detail.
```
aloukina:
```suggestion
This is a file in ``.json`` format that provides overall configuration options for an ``rsmeval`` experiment. Here's an `example configuration file <https://github.com/EducationalTestingService/rsmtool/blob/master/examples/rsmeval/config_rsmeval.json>`_ for ``rsmeval``. To make it easy to get started with  ``rsmeval``, we provide a way to automatically generate a configuration file that you can then edit based on your data and your needs. To do so, simply run ``rsmeval generate`` at the commmand line. If you have :ref:`subgroups <subgroups_eval>` in your data that you want to include in your analyses, run ``rsmeval generate --subgroups`` instead. Next, we describe all of the ``rsmeval`` configuration fields in detail.
```

doc/config_rsmpredict.rst
```diff
@@ -2,7 +2,7 @@
 
 Experiment configuration file
 """""""""""""""""""""""""""""
-This is a file in ``.json`` format that provides overall configuration options for an ``rsmpredict`` experiment. Here's an example configuration file for `rsmpredict <https://github.com/EducationalTestingService/rsmtool/blob/master/examples/rsmpredict/config_rsmpredict.json>`_.
+This is a file in ``.json`` format that provides overall configuration options for an ``rsmpredict`` experiment. Here's an `example configuration file <https://github.com/EducationalTestingService/rsmtool/blob/master/examples/rsmpredict/config_rsmpredict.json>`_ for ``rsmpredict``. To make it easy to get started with  ``rsmpredict``, we provide a way to automatically generate a configuration file that you can then just edit based on your data and your needs. To do so, simply run ``rsmpredict generate`` at the commmand line. Next, we describe all of the ``rsmpredict`` configuration fields in detail.
```
aloukina:
```suggestion
This is a file in ``.json`` format that provides overall configuration options for an ``rsmpredict`` experiment. Here's an `example configuration file <https://github.com/EducationalTestingService/rsmtool/blob/master/examples/rsmpredict/config_rsmpredict.json>`_ for ``rsmpredict``. To make it easy to get started with  ``rsmpredict``, we provide a way to automatically generate a configuration file that you can then edit based on your data and your needs. To do so, simply run ``rsmpredict generate`` at the commmand line. Next, we describe all of the ``rsmpredict`` configuration fields in detail.
```

doc/config_rsmsummarize.rst
```diff
@@ -3,7 +3,7 @@
 Experiment configuration file
 """""""""""""""""""""""""""""
 
-This is a file in ``.json`` format that provides overall configuration options for an ``rsmsummarize`` experiment. Here's an example configuration file for `rsmsummarize <https://github.com/EducationalTestingService/rsmtool/blob/master/examples/rsmsummarize/config_rsmsummarize.json>`_.
+This is a file in ``.json`` format that provides overall configuration options for an ``rsmsummarize`` experiment. Here's an `example configuration file <https://github.com/EducationalTestingService/rsmtool/blob/master/examples/rsmsummarize/config_rsmsummarize.json>`_ for ``rsmsummarize``. To make it easy to get started with  ``rsmsummarize``, we provide a way to automatically generate a configuration file that you can then just edit based on your data and your needs. To do so, simply run ``rsmsummarize generate`` at the commmand line. Next, we describe all of the ``rsmsummarize`` configuration fields in detail.
```
aloukina:
```suggestion
This is a file in ``.json`` format that provides overall configuration options for an ``rsmsummarize`` experiment. Here's an `example configuration file <https://github.com/EducationalTestingService/rsmtool/blob/master/examples/rsmsummarize/config_rsmsummarize.json>`_ for ``rsmsummarize``. To make it easy to get started with  ``rsmsummarize``, we provide a way to automatically generate a configuration file that you can then edit based on your data and your needs. To do so, simply run ``rsmsummarize generate`` at the commmand line. Next, we describe all of the ``rsmsummarize`` configuration fields in detail.
```

doc/config_rsmtool.rst
```diff
@@ -3,7 +3,7 @@
 Experiment configuration file
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-This is a file in ``.json`` format that provides overall configuration options for an ``rsmtool`` experiment. Here's an example configuration file for `rsmtool <https://github.com/EducationalTestingService/rsmtool/blob/master/examples/rsmtool/config_rsmtool.json>`_.
+This is a file in ``.json`` format that provides overall configuration options for an ``rsmtool`` experiment. Here's an `example configuration file <https://github.com/EducationalTestingService/rsmtool/blob/master/examples/rsmtool/config_rsmtool.json>`_ for ``rsmtool``. To make it easy to get started with  ``rsmtool``, we provide a way to automatically generate a configuration file that you can then just edit based on your data and your needs. To do so, simply run ``rsmtool generate`` at the commmand line. If you know you have :ref:`subgroups <subgroups_rsmtool>` in your data that you want to include in your analyses, run ``rsmtool generate --groups`` instead. Next, we describe all of the ``rsmtool`` configuration fields in detail.
```
aloukina:
```suggestion
This is a file in ``.json`` format that provides overall configuration options for an ``rsmtool`` experiment. Here's an `example configuration file <https://github.com/EducationalTestingService/rsmtool/blob/master/examples/rsmtool/config_rsmtool.json>`_ for ``rsmtool``. To make it easy to get started with  ``rsmtool``, we provide a way to automatically generate a configuration file that you can then edit based on your data and your needs. To do so, simply run ``rsmtool generate`` at the commmand line. If you have :ref:`subgroups <subgroups_rsmtool>` in your data that you want to include in your analyses, run ``rsmtool generate --subgroups`` instead. Next, we describe all of the ``rsmtool`` configuration fields in detail.
```

rsmtool/rsmcompare.py
```diff
@@ -180,44 +175,59 @@ def main():
...
+        # run the experiment
+        logger.info('Output directory: {}'.format(args.output_dir))
+        run_comparison(abspath(args.config_file),
+                       abspath(args.output_dir))
 
-    # convert all paths to absolute to make sure
-    # all files can be found later
-    config_file = abspath(args.config_file)
-    output_dir = abspath(args.output_dir)
+    else:
```
aloukina: I think it should be `elif` with `else` raising an error for an unknown command?

rsmtool/utils/commandline.py
```diff
@@ -0,0 +1,321 @@
...
+    If ``uses_output_directory`` is ``True``, an ``output_dir`` positional
+    argument will be added to the "run" subcommand parser.
+
+    If ``allows_overwriting`` is ``True``, an ``-f``/``--force``
+    optional argument will be added to the "run" subcommand parser.
+
+    The ``extra_run_options`` list should contain a list of ``CmdOption``
+    instances which are added to the "run" subcommand parser one by one.
+
+    If ``uses_subgroups`` is ``True``, a ``--groups`` optional
```
aloukina:
```suggestion
    If ``uses_subgroups`` is ``True``, a ``--subgroups`` optional
```

rsmtool/utils/commandline.py
```diff
@@ -0,0 +1,321 @@
...
+    allows_overwriting : bool, optional
+        Add the ``-f``/``-force_write`` optional argument to the "run" subcommand
+        parser. This argument allows the output for the respective
+        tool to be overwritten even if it already exists (file) or contains
+        output (directory).
+    extra_run_options : list, optional
+        Any additional options to be added to the "run" subcommand parser,
+        each specified as a ``CmdOption`` instance.
+    uses_subgroups : bool, optional
+        Add the ``--groups`` optional argument to the "generate" subcommand
```
aloukina:
```suggestion
        Add the ``--subgroups`` optional argument to the "generate" subcommand
```

rsmtool/utils/commandline.py
```diff
@@ -0,0 +1,321 @@
...
+                                            help=f"automatically generate an "
+                                                 f"{name} configuration file")
+    parser_run = subparsers.add_parser('run',
+                                       help=f"run an {name} experiment")
+
+    ###################################################
+    # Setting up options for the "generate" subparser #
+    ###################################################
+    if uses_subgroups:
+        parser_generate.add_argument('--groups',
```
aloukina:
```suggestion
        parser_generate.add_argument('--subgroups',
```

rsmtool/rsmcompare.py
desilinguist: The parser handles that automatically since the only valid subcommands are `run` and `generate`.

mulhod:
I think this was in the original issue, but I'm not sure if it was perhaps meant for a subsequent PR or not. I thought that if a user tried to run an automatically-generated config _without changing all placeholder values__, it would fail in a way that indicated that.

```
$ rsmeval generate --subgroups >test_auto_cfg_eval.json; rsmeval test_auto_cfg_eval.json test_auto_cfg_eval -f
WARNING: Automatically generated configuration files MUST be edited to add values for required fields and even for optional ones depending on your data.
Output directory: test_auto_cfg_eval
Saving configuration file.
  File "/Users/mmulholland/Documents/rsmtool_dev/rsmdev/bin/rsmeval", line 11, in <module>
    load_entry_point('rsmtool', 'console_scripts', 'rsmeval')()
  File "/Users/mmulholland/Documents/rsmtool_dev/rsmtool/rsmtool/rsmeval.py", line 264, in main
    overwrite_output=args.force_write)
  File "/Users/mmulholland/Documents/rsmtool_dev/rsmtool/rsmtool/rsmeval.py", line 111, in run_evaluation
    'not found.\n'.format(configuration['predictions_file']))
FileNotFoundError: Error: Predictions file ENTER_VALUE_HERE not found.
```

desilinguist:
> I think this was in the original issue, but I'm not sure if it was perhaps meant for a subsequent PR or not. I thought that if a user tried to run an automatically-generated config _without changing all placeholder values_, it would fail in a way that indicated that.
> 
> ```
Yes, I thought about doing that but then I looked at the error message running with that automatically generated file produces and decided that's pretty self-explanatory and we don't need to add code for that. If you guys disagree, let me know.

mulhod:
@desilinguist Fair enough. It's maybe a nice-to-have maybe visually, but, yeah, probably not very helpful.

aloukina:
I agree that the message is self-explanatory enough, at least for now.

rsmtool/rsmcompare.py
```diff
@@ -180,44 +175,59 @@ def main():
...
+    # set up an argument parser via our helper function
+    parser = setup_rsmcmd_parser('rsmcompare',
+                                 uses_output_directory=True,
+                                 uses_subgroups=True)
+
+    # if the first argument is not one of the valid sub-commands
+    # or one of the valid optional arguments, then assume that they
+    # are arguments for the "run" sub-command. This allows the
+    # old style command-line invocations to work without modification.
+    if sys.argv[1] not in ['run',
```
mulhod: Nitpick: Maybe the list of available tasks (`run` and `generate`) can defined somewhere else and used here so that if some new task is added in the future, this list will automatically be up to date.

Aside from this (since I see the same block of code in each sub-command's `main` method, is it possible to extract it out to a function that can be used in each?

desilinguist: Yeah, I had also noticed the shared code and considered factoring it out but then I reconsidered because the code is quite short, pretty readable, and IMO it's nice to have this bit in the main script itself rather than in another function. For example, it's nice to know that `run_experiment()` is what's called in `rsmtool`, `run_evaluation()` in `rsmeval` etc. 

What do you think?

desilinguist: I do like the idea of a list containing the valid subcommands though. Let me do that.

rsmtool/configuration_parser.py
```diff
@@ -282,15 +282,21 @@ def __len__(self):
...
+        dictionary.
 
         Returns
         -------
-        config_names : str
-            A comma-separated list of names from the config dictionary.
+        config_string : str
+            A string representation of the underlying configuration
+            dictionary as encoded by ``json.dumps()``. Configuration
+            options meant for internal use are not included.
```
aloukina:
```suggestion
            dictionary as encoded by ``json.dumps()``.  The configuration
            dictionary only includes the configuration option 
            that can be set by the user. 
```

rsmtool/rsmeval.py
```diff
@@ -10,30 +10,34 @@
...
+from .utils.commandline import generate_configuration, setup_rsmcmd_parser
+from .utils.constants import VALID_PARSER_SUBCOMMANDS
 from .utils.logging import LogFormatter
 from .writer import DataWriter
 
 
-def run_evaluation(config_file_or_obj_or_dict, output_dir):
+def run_evaluation(config_file_or_obj_or_dict,
+                   output_dir,
+                   overwrite_output=False):
```
aloukina: Yikes. This default `False` setting will break a lot of my code since previously you didn't need to worry about overwriting when calling API. It probably makes sense to keep it the way you did it, but we'll want to make sure we document it: could you this to your PR description for when I am writing the release note?

tests/data/output/autogenerated_rsmeval_config.json
```diff
@@ -0,0 +1,34 @@
+{
+    // REQUIRED: replace "ENTER_VALUE_HERE" with the appropriate value!
```
aloukina: Shall we add a comment with `//[link to documentation]` ?

rsmtool/utils/commandline.py
```diff
@@ -0,0 +1,321 @@
...
+        if parser_option.nargs is not None:
+            argparse_option_kwargs['nargs'] = f"{parser_option.nargs}"
+
+        # add this argument to the parser
+        parser_run.add_argument(*argparse_option_args, **argparse_option_kwargs)
+
+    return parser
+
+
+def generate_configuration(name,
```
aloukina: Should this be 'context' for consistency with the rest of the code?

rsmtool/utils/commandline.py
```diff
@@ -0,0 +1,321 @@
...
+                           use_subgroups=False,
+                           as_string=False):
```
aloukina: Two suggestions: 
1. Can we make it return Configuration, str, or dict
When using API, I would most likely want to receive back dictionary rather than a configuration object. You can convert it using `to_dict` but I feel that returning a dictionary is more user friendly.
2. I'd like to have suppress_warnings argument for when one uses this in a notebook or a script?

rsmtool/rsmcompare.py
mulhod: What I had envisioned was probably a function that would take a function like `run_experiment` as one of its arguments or maybe something that would sort of prepare it to be called (so you would still see it being used, anyway). I'd have to look into it a bit more. But, nonetheless, I think it's fine the way it is.

desilinguist: Yeah, I know what you mean. I tried it after your comment here but it became too much of a meta-mess and not as readable honestly.

rsmtool/rsmeval.py
desilinguist: Updated the PR description.

rsmtool/utils/commandline.py
desilinguist: May be we should just make it so that the `Configuration` class is only used internally and any time we need to read or write configurations, we only use dictionaries?

tests/data/output/autogenerated_rsmeval_config.json
desilinguist: Done.

rsmtool/utils/commandline.py
desilinguist: Done.

rsmtool/utils/commandline.py
desilinguist: @aloukina I have now added the `suppress_warnings` argument to `generate_configuration()` and added a `--quiet` / `-q` switch for the `generate` subcommand. Note that since the default should always be to show the warning, both of these are false by default.

aloukina: The argument in favor of keeping Configuration object available to power users is the .filepath attribute. If you pass a dictionary, you cannot set custom working directory.

desilinguist: Okay, `generate_configuration()` now returns a dictionary instead of a `Configuration` object.

PR #402, opened 2020-03-17
Fix CmdOption on Python 3.6

### LINKED ISSUE #401: `CmdOption` broken on Python 3.6
I just noticed that the ReadTheDocs build is failing and realized it’s because I am using a python 3.7-only keyword argument when defining the `rsmtool.utils.commandline.CmdOption` named tuple. ReadTheDocs uses python 3.6 to build the docs.

There are two solutions:
1. We drop support for python 3.6 with RSMTool 8 - this is not as radical as it sounds as python 3.8.2 is already out and stable and Python 3.9.0 will be out by the time we release RSMTool 8 (it’s already on its 4th alpha).
2. We implement a fix for Python 3.6 - it’s a little hacky but it’s only a single line of code. I just implemented it and it works.

Thoughts @aloukina @mulhod? @aoifecahill also tagging you because you might have some thoughts about what dropping python 3.6 means downstream.

desilinguist:
This PR fixes #401. 

It does so by adding a workaround for the `defaults` keyword argument for `collections.namedtuple` that is only supported in Python 3.7+.  It also replaces the `capture_output` keyword argument for `subprocess.run()` calls that is used in `test_cli.py`.

It also make sure that Travis CI builds use Python 3.6 and that Azure ones use Python 3.8. This ensures that we are covering the two Python versions most important to us - 3.6 for internal use and the latest (3.8) for external use.

*Update*: 3.8 support is broken right now (see #403). So, we have to make do with Python 3.7 instead of 3.8 in Azure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ before_install:
 - ${HOME}/miniconda3/bin/conda update conda --yes
 
 install:
-- ${HOME}/miniconda3/bin/conda create --name rsmenv -c conda-forge -c defaults -c ets --file requirements.txt --yes --quiet
+- ${HOME}/miniconda3/bin/conda create --name rsmenv -c conda-forge -c defaults -c ets --file requirements.txt python=${TRAVIS_PYTHON_VERSION} --yes --quiet
 - ${HOME}/miniconda3/envs/rsmenv/bin/pip install nose-cov
 - ${HOME}/miniconda3/envs/rsmenv/bin/pip install -e .
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,6 +5,7 @@
 
 variables:
   MPLBACKEND: Agg
+  PYTHON_VERSION: 3.8
 
 trigger:
   branches:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,7 +5,7 @@
 
 variables:
   MPLBACKEND: Agg
-  PYTHON_VERSION: 3.8
+  PYTHON_VERSION: 3.7
 
 trigger:
   branches:
--- a/rsmtool/utils/commandline.py
+++ b/rsmtool/utils/commandline.py
@@ -30,9 +30,13 @@ from .constants import CHECK_FIELDS, CONFIGURATION_DOCUMENTATION_SLUGS, DEFAULTS
 # will default to `None`.
 CmdOption = namedtuple('CmdOption',
                        ['dest', 'help', 'shortname', 'longname', 'action',
-                        'default', 'required', 'nargs'],
-                       defaults=[None, None, None, None, None, None])
-
+                        'default', 'required', 'nargs'])
+# if rsmtool were python 3.7+ only, we could just use the `defaults`
+# keyword argument to specify the default values; but to support python
+# 3.6, we need to mess with the `__new__` constructor.
+# Adapted from: https://stackoverflow.com/a/18348004
+# TODO: replace this with `defaults` when we drop support for python 3.7
+CmdOption.__new__.__defaults__ = (None,) * 6
 
 def setup_rsmcmd_parser(name,
                         uses_output_directory=True,
--- a/rsmtool/utils/commandline.py
+++ b/rsmtool/utils/commandline.py
@@ -35,8 +35,9 @@ CmdOption = namedtuple('CmdOption',
 # keyword argument to specify the default values; but to support python
 # 3.6, we need to mess with the `__new__` constructor.
 # Adapted from: https://stackoverflow.com/a/18348004
-# TODO: replace this with `defaults` when we drop support for python 3.7
+# TODO: replace this with `defaults` when we drop support for python 3.6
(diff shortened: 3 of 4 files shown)


mulhod:
I assume this isn't supposed to happen?
```
(/Users/mmulholland/Documents/rsmtool_dev/rsmdev) ➜  rsmtool git:(401-fix-cmdoption-on-python36) ✗ python --version
Python 3.6.10
(/Users/mmulholland/Documents/rsmtool_dev/rsmdev) ➜  rsmtool git:(401-fix-cmdoption-on-python36) ✗ nosetests -x --nologcapture tests
...................E
======================================================================
ERROR: test_cli.TestToolCLI.test_default_subcommand_is_run('rsmtool', '/Users/mmulholland/Documents/rsmtool_dev/rsmtool/tests/data/experiments/lr/lr.json /var/folders/fz/2bc74hd90qx1m748cffbxvkxvw9dhw/T/tmpbh2w_qqz', '/var/folders/fz/2bc74hd90qx1m748cffbxvkxvw9dhw/T/tmpbh2w_qqz', None)
----------------------------------------------------------------------
  File "/Users/mmulholland/Documents/rsmtool_dev/rsmdev/lib/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/mmulholland/Documents/rsmtool_dev/rsmtool/tests/test_cli.py", line 164, in check_tool_cmd
    encoding='utf-8')
  File "/Users/mmulholland/Documents/rsmtool_dev/rsmdev/lib/python3.6/subprocess.py", line 423, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/Users/mmulholland/Documents/rsmtool_dev/rsmdev/lib/python3.6/subprocess.py", line 729, in __init__
    restore_signals, start_new_session)
  File "/Users/mmulholland/Documents/rsmtool_dev/rsmdev/lib/python3.6/subprocess.py", line 1364, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'rsmtool': 'rsmtool'

----------------------------------------------------------------------
Ran 20 tests in 1.932s

FAILED (errors=1)
Exception ignored in: <generator object TestLoader.loadTestsFromGeneratorMethod.<locals>.generate at 0x12f00ad00>
RuntimeError: generator ignored GeneratorExit
```

desilinguist:
No that should not happen. Hmm, I actually do remember running into this. I think this only happens if you didn't `pip install -e .` which means that the command-line entry point `rsmtool` is unavailable.

mulhod:
It was a false alarm. Sorry. I think it was related to me downgrading Python in the same environment. However, there could be other issues. For example, now I see this (with a new env/Python 3.6.10):
```
(/Users/mmulholland/Documents/rsmtool_dev/rsmdev) ➜  rsmtool git:(401-fix-cmdoption-on-python36) ✗ nosetests -x --nologcapture tests
........................E
======================================================================
ERROR: test_cli.TestToolCLI.test_generate('rsmtool', 'generate', None, None)
----------------------------------------------------------------------
  File "/Users/mmulholland/Documents/rsmtool_dev/rsmdev/lib/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/mmulholland/Documents/rsmtool_dev/rsmtool/tests/test_cli.py", line 175, in check_tool_cmd
    encoding='utf-8')
  File "/Users/mmulholland/Documents/rsmtool_dev/rsmdev/lib/python3.6/subprocess.py", line 423, in run
    with Popen(*popenargs, **kwargs) as process:
TypeError: __init__() got an unexpected keyword argument 'capture_output'

----------------------------------------------------------------------
Ran 25 tests in 54.432s

FAILED (errors=1)
Exception ignored in: <generator object TestLoader.loadTestsFromGeneratorMethod.<locals>.generate at 0x13773bf68>
RuntimeError: generator ignored GeneratorExit
```

desilinguist:
Ah, is `capture_output` also python 3.7 only?

mulhod:
Yeah, it seems to have been added only in 3.7. https://docs.python.org/3/library/subprocess.html

desilinguist:
🤦‍♂

mulhod:
So can we just use `check_output` then? That is my go-to, actually. Why was this other way used here?

desilinguist:
`subprocess.run()` is the recommended way to use `subprocess()` unless compatibility with older Python versions is needed. I can easily fix this. Give me a minute.

desilinguist:
If Python 3.6 is actually important, perhaps we should switch to 3.6 (instead of 3.7) for `.travis.yml` and keep using Python 3.8 for Azure?

mulhod:
That seems like a good way of covering the versions for now. Maybe 3.6 can be dropped soon anyway as you mentioned originally.

aloukina:
I don't have a good feel on how many people use 3.6. If supporting it is not a trivial single-line fix, then may be moving to 3.7 is the right thing to do?

Issue #397, opened 2020-03-09
Trim imports in `__init__.py`


desilinguist:
It looks like most people only import the 5 major functions (`run_experiment`, `run_evaluation`, etc.) directly from `rsmtool` so we can probably get rid of many of the other imports in `__init__.py` and make things go faster.

PR #409, opened 2020-03-19
Add interactive configuration generation

### LINKED ISSUE #354: Add config generator

desilinguist:
This PR closes #354 and #397.

It adds interactive configuration generation for all of the tools. Non-interactive or batch-mode configuration generation was previously added by #398. 

To try out interactive generation, just run `X generate --interactive` where `X` is any of the five RSM command line tools. You can also add `--subgroups` for X = `rsmtool`, `rsmeval`, and `rsmcompare`.

The actual interaction bit is mostly handled by the excellent [`prompt_toolkit`](https://python-prompt-toolkit.readthedocs.io/en/master/index.html) library. Here's what this this PR does specifically: 

- The configuration generation (both non-interactive and interactive) is handled by a class called `rsmtool.utils.commandline.ConfigurationGenerator`. The `generate()` method handles the non-interactive generation and the `interact()` method handles the interactive generation.

- All of the required fields for the 5 tools and _some_ of the more important optional fields are displayed to the user to provide inputs for them. The appropriate auto-completion and value validation are attached to each fields so that the user is shown appropriate options just like in interactive IPython sessions (which is where `prompt_toolkit` was spun out of). All of this is handled by the new class `rsmtool.utils.commandline.InteractiveField`. Essentially, when the `interact()` method is called, an `InteractiveField` instance is created for each of the chosen fields with the appropriate completer and validator attached and shown to the user. The values input by the user and then used to create a configuration dictionary and then eventually output as a string.

- The chosen fields are contained in `rsmtool.utils.constants.INTERACTIVE_MODE_METADATA` which also defines the metadata for each field. This metadata is used to structure the interaction model (what to display, how to complete, how to validate etc.). The metadata consists of: 
  - the label to be shown for the field in interactive mode, 
  - the type of the field
  - whether the field takes a single value or multiple values (think "subgroups")
  - whether the field takes its value from a fixed list of choices

- Testing for this specific functionality was quite tricky. We don't really want to test the `prompt_toolkit` functionality as that is already tested outside of our code. So, in keeping with the best practices of unit testing, we use mocking to mock the functionality provided by `prompt_toolkit` and only test that our code is doing the right thing. The new testing classes added here are `test_utils.TestInteractiveField` and `test_utils.TestInteractiveGenerate`. The first class checks that `InteractiveField` works as expected and the second checks that the `ConfigurationGenerator.interact()` works as expected. Between the two of them, we are covering the whole functionality.

- This PR also adds documentation as follows: it adds a main page at the top level called "Auto-generating configuration files" that contains documentation for interactive,  non-interactive generation (which was previously in each tool's configuration file page), and also API usage. I also added multiple screencasts that illustrate how the interactive generation works. Finally, I added links to this top-level page from the configuration file pages of all 5 tools and also all 5 tutorials. I updated the API documentation to include the documentation for the `ConfigurationGenerator` class and its `generate()` method and pointed to that from the top-level page too.

- This PR also removes all imports from `__init__.py` other than the 5 main API functions.  This meant minor changes to the API documentation.


aloukina:
When generating file for RSMTool, you get some of the optional fields (e.g. second human score, min, max). In the generated file though these are listed under `` // OPTIONAL: replace default values below based on your data.``  with some fields having indeed default values but other fields having the values the user had set in the interactive mode. Is there a way to regroup the fields to keep the unassigned fields together?

aloukina:
`rsmeval generate` displays the following prompt: ` Name of column in evaluation data that contains human scores, if any:` - this is not an optional field so it should be `Name of column in evaluation data that contains human scores` with `sc1` as default, right?

rsmtool/utils/constants.py
```diff
@@ -201,3 +205,69 @@
...
+                                                         'of full-sized images? (true/<u>false</u>)',
+                                                'type': 'boolean'},
+                             'train_label_column': {'label': 'Name of column in training data that '
+                                                             'contains human scores (<u>sc1</u>)'},
+                             'test_label_column': {'label': 'Name of column in evaluation data that '
+                                                            'contains human scores (<u>sc1</u>)'},
+                             'length_column': {'label': 'Name of column in training/evaluation data '
+                                                        'that contains repsonse lengths, if any'},
+                             'human_score_column': {'label': 'Name of column in evaluation data '
+                                                             'that contains human scores, if any'},
```
aloukina:
```suggestion
                                                             'that contains human scores (optional for RSMPredict'},
```

rsmtool/utils/constants.py
```diff
@@ -201,3 +205,69 @@
...
+                             'test_file': {'label': 'Path to evaluation data file',
+                                           'type': 'file'},
+                             'predictions_file': {'label': 'Path to file containing predictions',
+                                                  'type': 'file'},
+                             'system_score_column': {'label': 'Name of column containing predictions'},
+                             'trim_min': {'label': 'The lowest possible score',
+                                          'type': 'integer'},
+                             'trim_max': {'label': 'The highest possible score',
+                                          'type': 'integer'},
+                             'experiment_dir': {'label': 'Path to directory containing RSMTool experiment',
```
aloukina:
```suggestion
                             'experiment_dir': {'label': 'Path to the directory containing RSMTool experiment',
```

rsmtool/utils/constants.py
```diff
@@ -201,3 +205,69 @@
...
+                             'test_label_column': {'label': 'Name of column in evaluation data that '
+                                                            'contains human scores (<u>sc1</u>)'},
+                             'length_column': {'label': 'Name of column in training/evaluation data '
+                                                        'that contains repsonse lengths, if any'},
+                             'human_score_column': {'label': 'Name of column in evaluation data '
+                                                             'that contains human scores, if any'},
+                             'second_human_score_column': {'label': 'Name of column in evaluation '
+                                                                    'data that contains scores from '
+                                                                    'a second human, if any'},
+                             'exclude_zero_scores': {'label': 'Keep responses with scores of 0 in '
```
aloukina:
```suggestion
                             'exclude_zero_scores': {'label': 'Keep responses with human scores of 0 in '
```

rsmtool/utils/constants.py
```diff
@@ -201,3 +205,69 @@
...
+                                       'type': 'choice',
+                                       'choices': sorted(set(BUILTIN_MODELS + VALID_SKLL_MODELS))},
+                             'train_file': {'label': 'Path to training data file',
+                                            'type': 'file'},
+                             'test_file': {'label': 'Path to evaluation data file',
+                                           'type': 'file'},
+                             'predictions_file': {'label': 'Path to file containing predictions',
+                                                  'type': 'file'},
+                             'system_score_column': {'label': 'Name of column containing predictions'},
+                             'trim_min': {'label': 'The lowest possible score',
```
aloukina:
```suggestion
                             'trim_min': {'label': 'The lowest possible human score',
```

desilinguist:
> `rsmeval generate` displays the following prompt: ` Name of column in evaluation data that contains human scores, if any:` - this is not an optional field so it should be `Name of column in evaluation data that contains human scores` with `sc1` as default, right?

`human_score_column` is listed in the optional section for `rsmeval`. See here: https://github.com/EducationalTestingService/rsmtool/blob/master/rsmtool/utils/constants.py#L136

rsmtool/utils/constants.py
```diff
@@ -201,3 +205,69 @@
...
+                                          'type': 'integer'},
+                             'trim_max': {'label': 'The highest possible score',
```
aloukina:
```suggestion
                             'trim_max': {'label': 'The highest possible human score',
```

aloukina:
That's because it defaults to `sc1`... Our definition of optional is not very transparent as I am realizing.

desilinguist:
Ah, I see. Yes "sc1" will be automatically in the config file so you are saying I just need to display "sc1" in the interactive prompt.

desilinguist:
> When generating file for RSMTool, you get some of the optional fields (e.g. second human score, min, max). In the generated file though these are listed under ` // OPTIONAL: replace default values below based on your data.` with some fields having indeed default values but other fields having the values the user had set in the interactive mode. Is there a way to regroup the fields to keep the unassigned fields together?

This may not be possible because all of the optional fields are generated in sorted order right now and the comment is inserted before the first optional field. Doing this would require extra bookkeeping - perhaps we can just tweak the message in the comment?

aloukina:
I agree that it's a nice to have - let's see how much people use the functionality.

doc/automated_configuration.rst
```diff
@@ -0,0 +1,132 @@
...
+
+
+The configuration file ``example_rsmtool.json`` generated via the session is shown below:
+
+.. literalinclude:: assets/example_rsmtool.json
+    :language: javascript
+
+.. note:: Although we use ``rsmtool`` in the example above, the same instructions apply to all 5 tools; simply replace ``rsmtool`` with ``rsmeval``, ``rsmcompare``, etc.
+
+There are some configuration options that can accept multiple inputs. For example, the ``experiment_dirs`` option for ``rsmsummrize`` takes a list of ``rsmtool`` experiment directories for a summary report. These options are handled differently in interactive mode. To illusrate this, let's generate a configuration file for ``rsmsummrize`` by using the following command:
```
mulhod:
```suggestion
There are some configuration options that can accept multiple inputs. For example, the ``experiment_dirs`` option for ``rsmsummarize`` takes a list of ``rsmtool`` experiment directories for a summary report. These options are handled differently in interactive mode. To illustrate this, let's generate a configuration file for ``rsmsummarize`` by using the following command:
```

doc/automated_configuration.rst
```diff
@@ -0,0 +1,132 @@
...
+
+    rsmsummarize generate --interactive > example_rsmsummarize.json
+
+The following screencast shows the interactive session (click to play):
+
+.. raw:: html
+    
+   <script id="asciicast-313149" src="https://asciinema.org/a/313149.js" data-autoplay="false" async></script>
+
+And here is the generated configuration file for ``rsmsummrize``:
```
mulhod:
```suggestion
And here is the generated configuration file for ``rsmsummarize``:
```

doc/automated_configuration.rst
```diff
@@ -0,0 +1,132 @@
...
+
+We end with a list of important things to note about interactive generation:
+
+- Carefully read the instructions and notes displayed at the top when you first enter interactive mode.
+
+- If you do not redirect the output of the command to a file, the generated configuration file will simply be printed out.
+
+- You may see messages like "invalid option", "invalid file" on the bottom left while you are entering the value for a field. This is an artifact of real-time validation. For example, when choosing a training file for ``rsmtool``, the message "invalid file" may be displayed while you navigate to the actual file. Once you get to a valid file, this message should disappear.
```
mulhod:
```suggestion
- You may see messages like "invalid option" and "invalid file" on the bottom left while you are entering the value for a field. This is an artifact of real-time validation. For example, when choosing a training file for ``rsmtool``, the message "invalid file" may be displayed while you navigate to the actual file. Once you get to a valid file, this message should disappear.
```

doc/automated_configuration.rst
```diff
@@ -0,0 +1,132 @@
...
+
+- Required fields will *not* accept a blank input (just pressing enter) and show an error message in the bottom left until a valid input is provided.
```
mulhod:
```suggestion
- Required fields will *not* accept a blank input (just pressing enter) and will show an error message in the bottom left until a valid input is provided.
```

doc/automated_configuration.rst
```diff
@@ -0,0 +1,132 @@
...
+For more advanced or experienced users who want to quickly get started with a dummy configuration file
+that they feel comfortable editing manually, RSMTool also provides the capability to generate configuration files non-interactively. To do so, simply omit the ``--interactive`` switch in the commands above. For example, to generate a dummy configuration file for ``rsmtool``, the command to run would be:
+
+.. code-block:: bash
+
+    rsmtool generate > dummy_rsmtool.json
+
+When running this command, the following warning would be printed out to stderr:
+
+.. code-block:: bash
```
mulhod: It's not a huge deal, but things like `for` are being highlighted since this is a "bash" code block. It there any way to keep it blocked off this way but to _not_ do the highlighting?

doc/automated_configuration.rst
```diff
@@ -0,0 +1,132 @@
...
+
+    WARNING: Automatically generated configuration files MUST be edited to add values
+    for required fields and even for optional ones depending on your data
+
+This warning explains that the generated file *cannot* be used directly as input to ``rsmtool`` since the required fields are filled with dummy values. This can be confirmed by looking at the configuration file the command generates:
+
+.. literalinclude:: assets/dummy_rsmtool.json
+    :language: javascript
+
+Note the two comments demarcating where the required and the optional fields. Note also that the required fields are filled with the dummy value "ENTER_VALUE_HERE" that *must* be manually edited by the user. The optional fields are filled with default values that may also need to be further edited depending on the data being used.
```
mulhod:
```suggestion
Note the two comments demarcating the locations of the required and optional fields. Note also that the required fields are filled with the dummy value "ENTER_VALUE_HERE" that *must* be manually edited by the user. The optional fields are filled with default values that may also need to be further edited depending on the data being used.
```

doc/automated_configuration.rst
```diff
@@ -0,0 +1,132 @@
...
+
+Just like interactive generation, non-interactive generation is supported by all 5 tools: ``rsmtool``, ``rsmeval``, ``rsmcompare``, ``rsmpredict``, and ``rsmsummarize``. 
+
+Similarly, to include subgroup information in the reports for ``rsmtool``, ``rsmeval``, and ``rsmcompare``, just add ``--subgroups`` to the command. Note that unlike in interactive mode, this would *only* add subgroup-based sections to the ``general_sections`` list in the output file. You will need to manually edit the ``subgroups`` option in the configuration file to enter the subgroup column names.
+
+Generation API
+~~~~~~~~~~~~~~
+Interactive generation is only meant for end users and can only be used via the 5 command-line tools ``rsmtool``, ``rsmeval``, ``rsmcompare``, ``rsmpredict`` and ``rsmsummarize``. It cannot be used via the RSMTool API.
+
+However, the non-interactive generation can indeed by used via the API which can be useful for more advanced RSMTool users. To illustrate, here's some example Python code to generate a configuration for ``rsmtool`` in the form of a dictionary:
```
mulhod:
```suggestion
However, the non-interactive generation can indeed be used via the API which can be useful for more advanced RSMTool users. To illustrate, here's some example Python code to generate a configuration for ``rsmtool`` in the form of a dictionary:
```

rsmtool/utils/commandline.py
```diff
@@ -230,117 +253,620 @@ def existing_configuration_file(string):
...
     """
-    Automatically generate an example configuration for a given
-    command-line tool.
+    Class that encapsulates a configuration field that is computed interactively.
 
-    Parameters
+    Attributes
+    ----------
+    choices : list
+        List of possible choices for the field value if ``data_type`` is "choice".
```
mulhod: Just for consistency
```suggestion
        List of possible choices for the field value if ``data_type`` is ``choice``.
```

rsmtool/utils/commandline.py
```diff
@@ -230,117 +253,620 @@ def existing_configuration_file(string):
...
+        An empty list for all other types of fields.
+    complete_style : prompt_toolkit.shortcuts.prompt.CompleteStyle
+        A CompleteStyle that defines how to style the completions.
+        Set to ``CompleteStyle.MULTI_COLUMN`` for fields that have ``data_type``
+        of "choice". Set to ``None`` for all other field types.
```
mulhod:
```suggestion
        of ``choice``. Set to ``None`` for all other field types.
```

rsmtool/utils/commandline.py
```diff
@@ -230,117 +253,620 @@ def existing_configuration_file(string):
...
+    completer : prompt_toolkit.completions.base.Completer
+        A ``Completer`` object used to provide auto-completion for the field.
+        The actual completer used depends on the ``data_type`` of the field.
+        For example, for fields of type "choice", we use a ``FuzzyWordCompleter``,
```
mulhod:
```suggestion
        For example, for fields of type ``choice``, we use a ``FuzzyWordCompleter``,
```

Now that I'm thinking about it actually, shouldn't it be ``"choice"``? It's the raw value that is within \`\`. And I think variables can be simply represented with one \`. But, it's not that important, anyway. Just thought I'd point it out. There are other instances in this file where the style could be enforced, but I won't point them all out.

rsmtool/utils/commandline.py
```diff
@@ -230,117 +253,620 @@ def existing_configuration_file(string):
...
+        - "file" : a field that accepts a path to a file
+        - "format" : a field that accepts possible file formats (csv/tsv/xlsx)
+        - "id" : a field that accepts experiment/comarison/summary IDs
+        - "integer" : a field that accepts integer values
+        - "text" : a field that accepts open-ended text
+    label : str
+        The label for the field that will be displayed to the user.
+    prompt_method : callable
+        The function that will be used to compute the value for the field.
+        The main difference arises between field that accept only a single
```
mulhod:
```suggestion
        The main difference arises between fields that accept only a single
```

rsmtool/utils/commandline.py
```diff
@@ -230,117 +253,620 @@ def existing_configuration_file(string):
...
+        A ``Completer`` object used to provide auto-completion for the field.
+        The actual completer used depends on the ``data_type`` of the field.
+        For example, for fields of type "choice", we use a ``FuzzyWordCompleter``,
+        and for fields of type ``dir`` and ``file``, we use a ``PathCompleter``.
+    count : str
+        An attribute indicating whether the field accepts a "single" value
+        or "multiple" values. For fields that require multiple values, e.g.
+        subgroups, a different strategy is used for interactive display.
+    data_type : str
+        The string indicating the data type of the field.
```
mulhod:
```suggestion
        A string indicating the data type of the field.
```

rsmtool/utils/commandline.py
```diff
@@ -230,117 +253,620 @@ def existing_configuration_file(string):
...
+        self.validator = None
+
+        # now override these attributes as necessary depending on field data types
+        if self.data_type == 'boolean':
+            allow_empty = field_type == 'optional'
+            self.completer = WordCompleter(['true', 'false'])
+            self.validator = self._make_boolean_validator(allow_empty=allow_empty)
+        elif self.data_type == 'choice':
+            if not self.choices:
+                raise(ValueError, f"invalid list of choices for {field_name}")
```
mulhod: I think this results in an exception:

```
In [1]: raise(ValueError, "hello")
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-711d55f1853f> in <module>
----> 1 raise(ValueError, "hello")

TypeError: exceptions must derive from BaseException
```

```suggestion
                raise ValueError(f"invalid list of choices for {field_name}")
```

rsmtool/utils/commandline.py
```diff
@@ -230,117 +253,620 @@ def existing_configuration_file(string):
...
+    def _make_file_completer(self):
+        """
+        Private method that creates a completer for a field
+        with ``data_type`` of "file".
+
+        Returns
+        -------
+        completer : prompt_toolkit.completion.base.Completer
+            A ``Completer`` instance that suggests directory names
+            and files with csv/tsv/xlsx extensions as potential
```
mulhod: Should we mention all extensions or maybe but an "etc." in here?

rsmtool/utils/commandline.py
```diff
@@ -230,117 +253,620 @@ def existing_configuration_file(string):
...
+    def _make_file_validator(self):
+        """
+        Private method that creates a validator for a field
+        with ``data_type`` of "file".
+
+        Returns
+        -------
+        validator : prompt_toolkit.validation.Validator
+            A ``Validator`` instance that makes sure that only
+            actually existing files with csv/tsv/xlsx extensions
```
mulhod: Same as above.

rsmtool/utils/commandline.py
```diff
@@ -230,117 +253,620 @@ def existing_configuration_file(string):
...
+    def _make_file_format_validator(self):
+        """
+        Private method that creates a validator for a field
+        with ``data_type`` of "format".
+
+        Returns
+        -------
+        validator : prompt_toolkit.validation.Validator
+            A ``Validator`` instance that makes sure that only
+            valid RSMTool extensions (csv/tsv/xlsx) and
```
mulhod: Same as above.

doc/automated_configuration.rst
desilinguist: Yes, good idea. I will do this.

rsmtool/utils/commandline.py
desilinguist: Yes you are right, it should be `"choice"`. Good catch. I will try and find all other instances an fix them too.

rsmtool/utils/commandline.py
desilinguist: Nice catch!

rsmtool/utils/commandline.py
desilinguist: I will do that.

desilinguist:
Thanks for the comprehensive review @mulhod! Much appreciated.

Issue #407, opened 2020-03-19
Difference between standard means computed incorrectly when user supplies `0` as population mean or sd.


aloukina:
This line:

rsmtool/rsmtool/utils/metrics.py

https://github.com/EducationalTestingService/rsmtool/blob/3d0c6916dc29037f7fda678ec85fff6224a444e5/rsmtool/utils/metrics.py#L359 is supposed to check for Nones yet it evaluates to True when one of the values is `0`. 

 if any(y_true_observed_population_params) and not all(y_true_observed_population_params): 
is supposed to check for Nones yet it evaluates to True when one of the values is 0.
The same is done through the function: the syntax used to check for None also returns True when the value is 0.

This had two consquences:
1. If only `mean`  OR `sd` is zero (but not both), the function would raise an error that one of the parameters is mising.
2.  If both `mean` or `sd` were zero, the function would re-compute the mean and sd using the data instead of using the user-supplied values.

aloukina:
This line: https://github.com/EducationalTestingService/rsmtool/blob/3d0c6916dc29037f7fda678ec85fff6224a444e5/rsmtool/utils/metrics.py#L359 is supposed to check for Nones yet it evaluates to True when one of the values is `0`.

Issue #406, opened 2020-03-18
Possible errors in docstring for `Analyzer.metrics_helper()`?


desilinguist:
It looks like there are some content errors in the [docstring](https://rsmtool.readthedocs.io/en/stable/api.html#rsmtool.analyzer.Analyzer.metrics_helper) for `Analyzer.metrics_helper()`. For example, there's a self-reference to `population_human_score_mn` in the description to `population_human_score_mn`. We should double check if there are additional errors.

PR #410, opened 2020-03-19
Changes to DSM computation

aloukina:
This PR addresses #406 and #407  : 

1. Fixed the inconsistencies in documentation string.
2. As described in #407, the function for computing difference of standardized means was treating `0` as `None` for population mean and sd. I fixed this and added tests.


tests/test_utils.py
```diff
@@ -423,6 +423,39 @@ def test_difference_of_standardized_means_with_no_population_info():
...
+
+def test_difference_of_standardized_means_zero_population_sd_human():
+    y_true, y_pred = (np.array([3, 5, 1, 2, 2, 3, 1, 4, 1, 2]),
+                      np.array([2, 1, 4, 1, 5, 2, 2, 2, 2, 2]))
+    expected = None
+    diff_std_means = difference_of_standardized_means(y_true, y_pred,
+                                                     population_y_pred_mn=2.44,
+                                                     population_y_pred_sd=0.54,
+                                                     population_y_true_observed_mn=2.44,
+                                                     population_y_true_observed_sd=0)
```
desilinguist:
```suggestion
                                                      population_y_pred_mn=2.44,
                                                      population_y_pred_sd=0.54,
                                                      population_y_true_observed_mn=2.44,
                                                      population_y_true_observed_sd=0)
```

tests/test_utils.py
```diff
@@ -423,6 +423,39 @@ def test_difference_of_standardized_means_with_no_population_info():
...
 
+def test_difference_of_standardized_means_zero_population_sd_pred():
+    y_true, y_pred = (np.array([3, 5, 1, 2, 2, 3, 1, 4, 1, 2]),
+                      np.array([2, 1, 4, 1, 5, 2, 2, 2, 2, 2]))
+    expected = None
+    diff_std_means = difference_of_standardized_means(y_true, y_pred,
+                                                     population_y_true_observed_mn=2.44,
+                                                     population_y_true_observed_sd=0.54,
+                                                     population_y_pred_mn=2.44,
+                                                     population_y_pred_sd=0)
```
desilinguist:
```suggestion
                                                      population_y_true_observed_mn=2.44,
                                                      population_y_true_observed_sd=0.54,
                                                      population_y_pred_mn=2.44,
                                                      population_y_pred_sd=0)
```

rsmtool/utils/metrics.py
```diff
@@ -356,32 +356,44 @@ def difference_of_standardized_means(y_true_observed,
...
         (population_y_pred_sd,
          population_y_pred_mn) = (np.std(y_pred, ddof=ddof),
                                   np.mean(y_pred))
 
+    # if any of the standard deviations equal zero
+    # raise a warning and return None
+    if population_y_pred_sd == 0 or population_y_true_observed_sd == 0:
+        warnings.warn("Population standard deviations for the computation of "
+                      "DSM is zero. No value will be computed.")
+        return None
```
desilinguist: Should we just raise an error here instead of returning `None`? How does a return value of `None` get handled in the report?

aloukina: I'd argue against it: we allow constant system scores for system evaluation so there is no reason to raise an error for this particular metrics given that we already return `None` for `R2` and `r` and other metrics that require non-zero variance. 

Let me add a regression test for this case though to make sure nothing in the report fails.

desilinguist: Sure, that makes sense. Thanks!

rsmtool/utils/metrics.py
```diff
@@ -356,32 +356,44 @@ def difference_of_standardized_means(y_true_observed,
...
         raise ValueError('You must pass both `population_y_pred_mn` and '
                          '`population_y_pred_sd` or neither.')
 
     warning_msg = ('You did not pass population mean and std. for `{}`; '
                    'thus, the calculated z-scores will be zero.')
 
     # if the population means and standard deviations were not provided, calculate from the data
-    if not population_y_true_observed_mn or not population_y_true_observed_sd:
+    # We only check for mean since the function requires
+    # both of these to be set of both to be None
```
mulhod:
```suggestion
    # both of these to be set or both to be None
```

rsmtool/utils/metrics.py
aloukina: The report correctly shows either `0` or `nan` depending on the metrics: e.g. for `R2` and `kappa` we report `0` and for correlation we report `nan`. 

Having said that, I had to switch `sd == 0` to `np.islose(sd, 0)` to avoid ghost DSM where there should be none.

rsmtool/analyzer.py
```diff
@@ -1198,6 +1199,17 @@ def compute_metrics_by_group(self,
...
+        # tell user to expect to see many warnings.
+        zero_sd_scores = [score for (score, sd) in population_sd_dict.items() if
+                          np.isclose(sd, 0, atol=1e-07)]
+        if len(zero_sd_scores) > 0:
+            warnings.warn("The standard deviation for {} scores "
+                          "is zero (all scores are the same). You "
+                          "will see multiple warnings about DSM computation "
+                          "since this metrics is computed separately for "
+                          "each subgroup.".format(', '.join(zero_sd_scores)))
+
```
desilinguist:
```suggestion
            warnings.warn("The standard deviation for {} scores "
                          "is zero (all values are the same). You "
                          "will see multiple warnings about DSM computation "
                          "since this metric is computed separately for "
                          "each subgroup.".format(', '.join(zero_sd_scores)))

```

PR #414, opened 2020-04-03
Update SKLL to v2.1

desilinguist:
This PR closes #404 by updating the SKLL version. This required changing the test data for a few SKLL-specific tests.

PR #415, opened 2020-04-03
Update Azure Pipelines to use Python 3.8

### LINKED ISSUE #405: Change Azure pipelines python version to 3.8 from 3.7
Once #403 and #404 are resolved, we should change the Python version in `azure-pipelines.yml` to 3.8 from 3.7.

desilinguist:
This PR closes #405. It should not be reviewed until #414 is merged.

While Travis CI tests on Python 3.6, it would be good to test on Python 3.8 too and we can use Azure for that.


desilinguist:
@mulhod and @aloukina any chance you can approve this too? Thanks!

PR #418, opened 2020-04-09
Allow passing in fixed parameters to SKLL models

### LINKED ISSUE #136: Pass fixed parameters to SKLL
For SKLL learners that can take in extra arguments via `fixed_parameters`, we currently don't have a way to pass those into RSMTool. Perhaps we can add a config option like `skll_fixed_parameters`? 

We already added `skll_objective`. We don't want to replicate all of SKLL's field here though. In the longer term, perhaps a better option would be to have something like `skll_args` and let that be a dictionary that the user can pass any SKLL fields in?

### LINKED ISSUE #416: Config normalization/processing/validation being called twice
We are unnecessarily calling the normalize/process/validate classmethods twice. We call them in the `ConfigurationParser.parse()`  method but they are also called when the `Configuration` object is initiated right after.

desilinguist:
This PR closes #136 (and also closes #416). 

- Add a new configuration option called `skll_fixed_parameters` to allow passing in fixed hyperparameters to SKLL models.
- Link it through the entire pipeline so that it gets passed to the modeler, the reporter, and also to the actual report.
- Reorganize the model characteristics reported for SKLL models into an easier-to-read section called "Characteristics".
- Fix the bug where the normalize/process/validate configuration workflow was being called twice from inside `ConfigurationParser.parse()`. This should speed things up marginally.
- Add 3 new tests for custom SKLL parameters and update the configuration generation expected output to have the new option.
- Document the `skll_fixed_parameters` option.


rsmtool/configuration_parser.py
```diff
@@ -1235,7 +1233,18 @@ def validate_config(cls, config, context='rsmtool'):
                     raise ValueError("Invalid SKLL objective. Please refer to the SKLL "
                                      "documentation and choose a valid tuning objective.")
 
-        # 9. Check that if we are running rsmtool to ask for
+        # 9. Check if that if "skll_fixed_parameters" is specified,
+        # it's specified for SKLL model and _not_ a built-in linear
+        # regression model; we cannot check whether the parameters
+        # are valid at parse time but SKLL will raise an error#
+        # at run time for any invalid parameters
+        if new_config['skll_fixed_parameters']:
```
aloukina, line 1241: Should this be an error to make sure the user doesn't ignore the warning?

desilinguist: We raise a warning for `skll_objective` too so I just followed the same logic here.

rsmtool/notebooks/skll_model.ipynb
```diff
@@ -14,29 +14,19 @@
...
-    "Markdown('Number of features in model: **{}**'.format(len(features_used)))"
+    "intro_markdown_str = f\"\"\"### Characteristics \n",
+    "\n",
+    "- Model: **{model_name}**\n",
+    "- Number of features: **{len(features_used)}**\n",
+    "- Tuning objective: **{objective_used}**\n",
+    "\"\"\"\n",
+    "\n",
+    "if skll_fixed_parameters:\n",
+    "    intro_markdown_str += f\"- Custom fixed parameters: **{skll_fixed_parameters}**\"\n",
```
aloukina, line 27: We *could* print the parameters as a bullet list to make the report look nicer but it is a nice to have.

desilinguist: I agree that would look nicer. Let me see if that's an easy fix.

rsmtool/configuration_parser.py
aloukina: Makes sense!

rsmtool/notebooks/skll_model.ipynb
desilinguist: Just pushed a commit to print as bulleted list. Try it out?

aloukina: Looks great to me - thank you!

rsmtool/configuration_parser.py
```diff
@@ -1235,7 +1233,18 @@ def validate_config(cls, config, context='rsmtool'):
                     raise ValueError("Invalid SKLL objective. Please refer to the SKLL "
                                      "documentation and choose a valid tuning objective.")
 
-        # 9. Check that if we are running rsmtool to ask for
+        # 9. Check if that if "skll_fixed_parameters" is specified,
```
mulhod, line 1236:
```suggestion
        # 9. Check that if "skll_fixed_parameters" is specified,
```

rsmtool/configuration_parser.py
```diff
@@ -1235,7 +1233,18 @@ def validate_config(cls, config, context='rsmtool'):
...
+        # it's specified for SKLL model and _not_ a built-in linear
+        # regression model; we cannot check whether the parameters
+        # are valid at parse time but SKLL will raise an error#
```
mulhod, line 1239:
```suggestion
        # are valid at parse time but SKLL will raise an error
```

rsmtool/modeler.py
```diff
@@ -1060,6 +1060,7 @@ def train_skll_model(self,
                          filedir,
                          figdir,
                          file_format='csv',
+                         custom_fixed_parameters={},
```
mulhod, line 1063: Nitpick: While probably not a big deal in this case, it might be better if the default argument did not have a mutable value because weird things could happen in the future if the parameter's value were to be mutated within this function. Maybe using `None` instead and then modifying where it's used to substitute in an empty dictionary if the value is `None` would be slightly more future-proof.

desilinguist: Good call! I will fix this.

Issue #392, opened 2020-03-05
Deprecate INI style configuration files


desilinguist:
The INI-style configuration files are not used internally at ETS and they don't seem to be any more useful compared to the JSON format. So, they will be deprecated.

Issue #368, opened 2020-02-14
Deprecate methods/attributes/functionality we promised to deprecate


aloukina:
We have a variety of deprecation warnings throughout the code.
We should actually deprecate them in the next major release.

aloukina:
The following methods are deprecated:

* Support for older experiments that use rounded scores for QWK in RSMCompare
* Support for older experiments use SMD for subgroups (instead of DSM).

PR #423, opened 2020-04-23
Deprecate what we promised to deprecate

### LINKED ISSUE #366: Deprecate get_trim_min_max
We are now displaying the following error:

`get_trim_min_max method has been replaced by  get_trim_min_max_tolerance() and will not be supported in future releases.`

We should actually deprecated it in RSMTool 8.0

aloukina:
This PR closes #366, #392 and #368 .
The following are deprecated:
* get_trim_max method
* ini/cfg configuration files
* Support for older experiments that use rounded scores for QWK in RSMCompare
* Support for older experiments that use SMD for subgroups (instead of DSM) in RSMCompare
* Support for old positional argument to Configuration()
* Support for Configuration.filepath() attribute
* Support for old config files (for R based RSMTool) (not used at least since RSMTool 5.5)
* Support for old model names (used in R-based RSMTool) (not used at least since RSMTool 5.5)
* Raising error when somebody requests empWtDropNeg
* Support for old feature.json file (not used at least since RSMTool 5.5)

I also updated RSMCompare and RSMSummarize tests since they used old column names :)


rsmtool/preprocessor.py
```diff
@@ -2446,7 +2386,7 @@ def process_data_rsmpredict(self, config_obj, data_container_obj):
 
         # if we are using a newly trained model, use trim_tolerance from the
```
aloukina, line 2387: I kept this as is because sometimes we want to use very old models to produce scores.

rsmtool/comparer.py
```diff
@@ -123,32 +123,15 @@ def _modify_eval_columns_to_ensure_version_compatibilty(df,
...
-                warnings.warn("Please note that newer versions of RSMTool (7.0 or greater) use only the trimmed "
-                              "scores for weighted kappa calculations. Comparisons with experiments using "
-                              "`trim_round` for weighted kappa calculations will be deprecated in the next "
-                              "major release.", category=DeprecationWarning)
+            raise ValueError("Please note that newer versions of RSMTool (7.0 or greater) use only "
+                             " the trimmed scores for weighted kappa calculations. At least one "
+                             "of your experiments was created "
+                             "using an older version of RSMTool and uses trim_round for  "
+                             "`trim_round` for weighted kappa calculations. Please re-run "
+                             "the experiments with the recent version of RSMTool.")
```
desilinguist, line 134: How about?

```suggestion
            raise ValueError("RSMTool (7.0 or greater) uses only "
                             "trimmed scores for weighted kappa calculations. At least one "
                             "of your experiments was run "
                             "using an older version of RSMTool that used  "
                             "trimmed and rounded scores instead. Please re-run "
                             "the experiment with the latest version of RSMTool.")
```

rsmtool/configuration_parser.py
```diff
@@ -195,8 +149,7 @@ def __init__(self,
         filename : str, optional, keyword-only
             The name of the configuration file.
             The file must be stored in configdir.
-            This argument is not used in RSMTool and only added for
-            backwards compatibility for the deprecated `filepath` attribute.
+            This argument is currently not used in RSMTool.
```
desilinguist, line 152: Can't we just remove it?

rsmtool/configuration_parser.py
```diff
@@ -1051,19 +876,7 @@ def normalize_config(cls, config):
...
-                                                    norm_field_name),
-                              category=DeprecationWarning)
-            else:
-                norm_field_name = field_name
-
-            new_config[norm_field_name] = config[field_name]
+            new_config[field_name] = config[field_name]
 
         # Convert old values for prediction scaling:
         if 'use_scaled_predictions' in new_config:
```
desilinguist, line 882: Can we deprecate the old values "scale" and "raw also?

rsmtool/configuration_parser.py
```diff
@@ -928,61 +808,6 @@ def _parse_json_file(self, filepath):
 
         return configdict
 
-    def _parse_cfg_file(self, filepath):
```
desilinguist, line 931: I still see a call to `_parse_cfg_file` in the if statement in `ConfigurationParser.parse()`. We should remove that.

rsmtool/comparer.py
desilinguist, line 134: How about?

```suggestion
            raise ValueError("RSMTool (7.0 or greater) uses only "
                             "trimmed scores for weighted kappa calculations. At least one "
                             "of your experiments was run "
                             "using an older version of RSMTool that used  "
                             "trimmed and rounded scores instead. Please re-run "
                             "the experiments with the latest version of RSMTool.")
```

rsmtool/configuration_parser.py
desilinguist, line 152: Can we just remove it then?

rsmtool/configuration_parser.py
desilinguist, line 882: Can we also deprecate the old values for scaling predictions: "scale" and "raw"?

rsmtool/comparer.py
aloukina: I replaced "trimmed" with "unrounded" since this is actually more accurate. Let me know if it's not ok.

aloukina: Same here.

rsmtool/configuration_parser.py
aloukina: I'd prefer to keep it in case `.filepath` is used in ETS module.

rsmtool/configuration_parser.py
aloukina: Good catch. We actually don't need the whole normalized method anymore!

tests/test_configuration_parser.py
```diff
@@ -35,83 +35,6 @@ class TestConfigurationParser:
...
-
-        configdir = Path(getcwd())
-        configuration = Configuration(data)
-        eq_(configuration._configdir, configdir)
-
-    @raises(TypeError)
-    def test_load_config_from_dict_wrong_type(self):
-        data = [('expID', 'test')]
-        configdir = 'some/config/dir'
-        Configuration(data, configdir)
```
desilinguist, line 114: Hmm, something I missed the last time around, why are we removing the test functions other than `test_normalize_config()`? Don't they still apply?

rsmtool/configuration_parser.py
```diff
@@ -1008,100 +801,15 @@ def parse(self, context='rsmtool'):
         """
         extension = Path(self._filename).suffix.lower()
```
desilinguist, line 802: We don't need this line anymore.

rsmtool/configuration_parser.py
```diff
@@ -166,19 +121,16 @@ class Configuration:
...
-                 filename=None,
                  context='rsmtool'):
         """
         Create an object of the `Configuration` class.
 
-        This method should be used to directly instantiate a Configuration
-        object only if you already have a normalized configuration dictionary
-        (e.g., from previous RSMTool experiments).
+        This method can be used to directly instantiate a Configuration
+        object.
```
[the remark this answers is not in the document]
aloukina: We now call `validate` and `process`  in the `init` function: see lines 157 and 158 below. We missed changing this text during previous PR review...

tests/test_configuration_parser.py
aloukina: I added back the test for TypeError. 

The other two are leftovers from earlier tests and implicitly test normalization (all field names are from the old R-style config file). We are already testing for the same functionality in `test_init_default_values` (tests for initiating from dictionary with no configdir), `test_init_with_configdir_only_as_kword_argument` (tests for initiating with configdir).

desilinguist: Ah, great. Makes sense. Thanks!

PR #427, opened 2020-05-01
Remove no longer needed warning filters

### LINKED ISSUE #331: Remove warning filters
[From JIRA] Currently, we have some warnings being filtered both in the header notebook and also in some of the actual python code files. Periodically, we should check to make sure that such filtering is actually needed. For example, filters for FutureWarnings may be removed once the modules have been updated.

See also #352 and #363

desilinguist:
This PR closes #331.

The only warning filters I removed for for `statsmodels` in `modeler.py` and in two RSMTool notebooks. I checked the Travis CI logs to confirm and no such warnings showed up.


aloukina:
Thank you for taking care of this!

PR #428, opened 2020-05-04
Improve check for linear model

### LINKED ISSUE #393: Make check in RSMTool more explicit
There's a currently [a place in RSMTool](https://github.com/EducationalTestingService/rsmtool/blob/master/rsmtool/rsmtool.py#L286) where we raise a `ValueError` if a coefficients file already exists in the output directory. There's no documentation as to why this is being checked.

From @aloukina, this is why we do this:

> This is what I think it’s doing in a roundabout way: it uses the existence of coefficients file as a shortcut to establish that we are dealing with a linear model.  If you first run a LR experiment and then followed it with SVR using the same id and the same output_directory, you’ll get this error. The coefficients file will be there but the modeler will not be able to scale them because [these lines](https://github.com/EducationalTestingService/rsmtool/blob/0eb9a96e236828483123e46a4b1e2f1376c59827/rsmtool/modeler.py#L1422) will fail. 

We should replace this roundabout check with a much more explicit check, e.g., simply testingwhether `predconfig.get_coefficients()` fails? That would make it much more readable.

desilinguist:
This PR fixes #393. 

- It tweaks `Modeler.scale_coefficients()`  to raise an exception when its call to `get_coefficients()` fails beacuse the model is non-linear, for example.

- It also tweaks the error message for the special scenario in `rsmtool.py` where the user previously ran the same experiment with the same ID with a linear model and then is running it again with a non-linear model, show a more descriptive error message.

- It adds a test for this special scenario.


rsmtool/rsmtool.py
```diff
@@ -269,27 +269,33 @@ def run_experiment(config_file_or_obj_or_dict,
...
 
-            # Scale coefficients, and return DataContainer w/ scaled coefficients
+            # scale coefficients, and return DataContainer w/ scaled coefficients
             scaled_data_container = modeler.scale_coefficients(pred_config)
 
-            # Write out files to disk
+        # unless no coefficients are available this time around which can
+        # happen if the user is re-running the same experiment with the same
+        # ID but with a non-linear model whereas the previous run of the same
+        # ID was with a linear model and the user has not cleared the directory
```
aloukina, line 284:
```suggestion
        # Raise an error if the coefficient file exists but the 
        # coefficients are not available for the current model 
        #  which can happen if the user is re-running the same experiment with the same
        # ID but with a non-linear model whereas the previous run of the same
        # ID was with a linear model and the user has not cleared the directory
```

PR #431, opened 2020-05-06
Tweak README file and setup.py

desilinguist:
- Update `README.rst` and `setup.py` to include Python 3.7 and 3.8.
- Add some text about interactive config generation video and link to the demo video.
- Add a getting started section heading to make things more readable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@
 __pycache__
 /rsmtool.sublime-workspace
 /rsmtool.sublime-project
+/.condaenv
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,3 @@
 *test_outputs/
 
 __pycache__
-/rsmtool.sublime-workspace
-/rsmtool.sublime-project
-/.condaenv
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -15,3 +15,4 @@
 __pycache__
 /rsmtool.sublime-workspace
 /rsmtool.sublime-project
+/.condaenv
```
mulhod, line 18: Nitpick: Some of these entries might be better-suited in `.git/info/exclude` (that is, custom repository-level ignores) since they are specific to how a particular user names a Sublime project (or even uses Sublime), manages Conda envs, etc. The last here might be a bit confusing to a new developer wondering where this file comes from. Anyway, just something to think about.

desilinguist: YesI That is a good call - developer-specific stuff should not go in `.gitignore`. Let me clean this up.

desilinguist:
Good catch! Will do.

doc/getting_started.rst
```diff
@@ -11,9 +11,9 @@ Currently, the recommended way to install RSMTool is by using the ``conda`` pack
 
 1. To install ``conda``, follow the instructions on `this page <https://conda.io/projects/conda/en/latest/user-guide/install/index.html>`_. 
 
-2. Create a new conda environment (say, ``rsmtool``) and install the RSMTool conda package by running::
+2. Create a new conda environment (say, ``rsmtool``) and install the RSMTool conda package for, say, Python 3.8 by running::
```
aloukina, line 14:
```suggestion
2. Create a new conda environment (say, ``rsmtool``) and install the RSMTool conda package using your preferred Python version. For example, for Python 3.8 run::
```

Issue #449, opened 2020-07-31
Handle `UserWarning` for feature distribution plots


desilinguist:
With new versions of `matplotlib`, we seem to get the following warning in the notebook:

```
UserWarning: FixedFormatter should only be used together with FixedLocator
```

Right now, we just wrap the whole plotting code in a `catch_warnings()` block but this will filter out all warnings. We should figure out the underlying issue and fix it.

Issue #447, opened 2020-07-31
DTypewarning that we use in the reader is no longer supported in recent Pandas


aloukina:
Pandas no longer support DTypeWarning so we get the following:

```packages\rsmtool\reader.py", line 263, in read_from_file
    warnings.filterwarnings('ignore', category=pd.io.common.DtypeWarning)
AttributeError: module 'pandas.io.common' has no attribute 'DtypeWarning'```

(And probably in other places too).

Here are some potentially related issues:

https://stackoverflow.com/questions/60199213/pandas-read-csv-error-due-to-pandas-io-common-not-importing-is-url-in-1-0-x
https://stackoverflow.com/questions/40816169/pandas-datareader-raises-attributeerror-module-pandas-io-has-no-attribute-da/43289270

desilinguist:
`DtypeWarning` has moved to `pandas.errors`.

Issue #446, opened 2020-07-31
Better handling of no arguments situation


aloukina:
If the user tries to run `rsmtool` with no arguments, they get an error from argument parser (list out of range). We should probably handle it more elegantly and in this case default to -h? Or at least show a helpful message?

desilinguist:
Yeah, this should be an easy fix to just check to add [here](https://github.com/EducationalTestingService/rsmtool/blob/master/rsmtool/rsmtool.py#L353) based on len(sys.argv).

PR #450, opened 2020-08-01
Handle no arguments to tools and fix code to handle pandas changes

desilinguist:
This PR addresses #446 and #447. 

- Handle no arguments being passed to all CLI tools and add corresponding test.
- Use correct module for `DtypeWarning` due to change in `pandas`. 
- Replace `check_less_precise` keyword argument in `assert_frame_equal()` to avoid warnings.
- Move full plotting block for training set feature distributions under a `catch_warnings()` to avoid a strange `UserWarning` (see #449).


rsmtool/reader.py
```diff
@@ -260,7 +260,7 @@ def read_from_file(filename, converters=None, **kwargs):
 
         # ignore warnings about mixed data types for large files
         with warnings.catch_warnings():
-            warnings.filterwarnings('ignore', category=pd.io.common.DtypeWarning)
```
mulhod, line 263: Would this change affect using earlier versions of `pandas`?

rsmtool/test_utils.py
```diff
@@ -127,7 +127,13 @@ def check_run_experiment(source,
     if consistency:
         check_consistency_files_exist(output_files, experiment_id, file_format=file_format)
 
-    check_report(html_report)
+    check_report(html_report, raise_warnings=False)
+
+    # we want to ignore deprecation warnings for RSMTool, so we remove
+    # them from the list; then, we make sure that there are no other warnings
+    warning_msgs = collect_warning_messages_from_report(html_report)
+    warning_msgs = [msg for msg in warning_msgs if 'DeprecationWarning' not in msg]
```
aloukina, line 135: Would this also ignore deprecation warnings from other packages though?

tests/test_cli.py
```diff
@@ -46,6 +46,36 @@ def tearDownClass(cls):
         for tempdir in cls.temporary_directories:
             tempdir.cleanup()
 
+    def check_no_args(self, context):
```
aloukina, line 49: Thank you :)

rsmtool/reader.py
desilinguist: Good catch! Yes, it will break for older versions of pandas. So we should pin pandas to be >=1.0.5 where this change was introduced?

rsmtool/test_utils.py
desilinguist: Yes it would.

rsmtool/reader.py
mulhod: Yeah, that would probably be the easiest way to handle this. But, if we want to allow for earlier versions of `pandas`, perhaps we could have a try/except at the top to import `DtypeWarning` from first `pandas.errors` and then, if that fails, `pandas.io.common` and then we could just use `DtypeWarning` as the category here.

desilinguist: I like that idea much better!

rsmtool/notebooks/header.ipynb
```diff
@@ -262,7 +262,7 @@
     "converter_dict = {column: str for column in string_columns}\n",
     "\n",
     "with warnings.catch_warnings():\n",
-    "    warnings.filterwarnings('ignore', category=pd.io.common.DtypeWarning)\n",
```
mulhod, line 265: The suggested change below would also apply here. I wonder if you could even import `DtypeWarning` from `rsmtool.reader`.

desilinguist: Yep, I did the conditional import here too.

PR #457, opened 2020-08-03
Add strict mode for tests.

### LINKED ISSUE #455: Add a STRICT_MODE environment variable for tests
Right now, we are ignoring deprecation warnings in notebook reports. By using this environment variable, we can make it so that these warnings lead to test failures instead. This is useful since this would allow us to run in strict mode locally to easily find any such warnings before a release.

desilinguist:
This PR closes #455 . 

It add support for an environment variable called `STRICT` which, if set, will _not_ ignore deprecation warnings in reports. This environment variable is only meant to be use for local test running and before a major release. I have also updated the release process document to indicate the same.

### DIFF
--- a/doc/internal/release_process.rst
+++ b/doc/internal/release_process.rst
@@ -3,11 +3,13 @@ RSMTool Release Process
 
 This process is only meant for the project administrators, not users and developers.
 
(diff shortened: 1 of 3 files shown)


aloukina:
Thank you for also updating the documentation!

PR #464, opened 2020-09-22
Hotfix for nbconvert issue & seaborn warning fix

### LINKED ISSUE #461: Stable branch breaks with latest version of nbconvert v6.0.2
Looks like they [broke something](https://github.com/jupyter/nbconvert/issues/1394) in the latest version of nbconvert that was released a few days ago. 

If they don't fix it in a day or two, we'll have to pin to an older version and release a hotfix. In the meanwhile, a temporary fix is to force downgrade `nbconvert` in the new environment like so:

```
conda install --no-deps nbconvert=5.5
```

desilinguist:
This PR closes #461 and #463. 

1. It pins `nbconvert` to `<6.0` since the new versions are broken.

2. It uses the proper signature for seaborn function calls to prevent FutureWarnings from appearing in the rendered notebooks.


mulhod (approved):

requirements.txt
```diff
@@ -2,6 +2,7 @@ doc2dash
 jupyter
 nose
 notebook
+nbconvert<6.0
```
aloukina, line 5: To remove if a new version is released.

aloukina (approved):
Thank you for taking care of this!

Issue #367, opened 2020-02-14
Consistent approach to warnings


aloukina:
In some parts of the code we use `logger.warning`while in other parts we used `warnings`.
Is there any reason not use `warnings` everywhere?

desilinguist:
This is actually addressed in the Python [logging HOWTO](https://docs.python.org/3/howto/logging.html): 

> warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning. 
> 
> logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted.

aloukina:
We have a few inconsistencies in how we apply this. For example we raise UserWarning when computing partial correlations.  Another concern I have is that on user side both types of warnings looks very similar so we are training the users to ignore all warnings.

desilinguist:
On a related note, we should be using `FutureWarning` for deprecation warnings and not `DeprecationWarning` since that's what the Python documentation [recommends](https://docs.python.org/3/library/exceptions.html#FutureWarning) and this is what scikit-learn has recently done.

desilinguist:
@aloukina this is done right?

aloukina:
No, not throughout the code ;(

PR #469, opened 2020-09-28
Fix pie chart and other minor changes

### LINKED ISSUE #460: Tweak warning
RSMTool issues the following warning if the `select_transformations` flag is set to True in config. However, it gives it even if a feature file is not specified and the list of features is directly fed into the experiment configuration.

> WARNING: You specified a feature file but also set `select_transformations` to True. Any transformations or signs specified in the feature file will be overwritten by the automatically selected transformations and signs.

https://github.com/EducationalTestingService/rsmtool/blob/27de5f63e78efe9d40ca3415513aab03c09f4eb9/rsmtool/configuration_parser.py#L997

### LINKED ISSUE #466: Document STRICT-MODE under testing documentation
We currently describe STRICT-MODE functionality under release process but I was actually looking for it under testing documentation. It would make sense to add it there too.

### LINKED ISSUE #468: Set explicit normalization  in pie chart.
We have a pie chart for relative coefficients. Currently normalization is set to None which raises a deprecation warning:

```<ipython-input-41-a9381304d58f>:19: MatplotlibDeprecationWarning: normalize=None does not normalize if the sum is less than 1 but this behavior is deprecated since 3.3 until two minor releases later. After the deprecation period the default value will be normalize=True. To prevent normalization pass normalize=False```

We need to explicitly set it to False to avoid normalization when we have negative coefficients. The goal here is to make it visually clear to the user the pie chart is not appropriate for the model.

### PR DESCRIPTION
**Pull request by aloukina**
This PR closes #468, closes #460 and closes #466

* The code for the relative coefficient pie chart now explicitly sets normalize to true. The chart is only displayed if there are no negative coefficients. 
* Updated documentation to explain strict mode in sections about tests.
* The warning about transformations is only shown if the user passed a feature file. It is suppressed if the user gave a list of features. 
* Added tests for the above. 
* PEP8 fixes to configuration_parser.py.

### CODE REVIEW

File: rsmtool/notebooks/builtin_model.ipynb [hunk 1]
```diff
@@ -151,11 +151,13 @@
     "    ax1.set_title('Values of standardized coefficients')\n",
     "    ax1.set_xlabel('')\n",
     "    ax1.set_ylabel('')\n",
-    "    # no pie chart if we have more than 15 features or if the feature names are long (pie chart looks ugly)\n",
-    "    if len(features_used) <= 15 and longest_feature_name <= 10:\n",
+    "    # no pie chart if we have more than 15 features,\n",
+    "    # if the feature names are long (pie chart looks ugly)\n",
+    "    # or if there are any negative coefficients.\n",
+    "    if len(features_used) <= 15 and longest_feature_name <= 10 and (df_betas_sorted['relative']>=0).all():\n",
```
desilinguist, line 157: Do we want a test for this?

desilinguist (changes requested):
Thanks for taking care of these! I have some suggestions and questions.

rsmtool/notebooks/builtin_model.ipynb
aloukina: We don't really have tests for plots in the notebooks... It would be good in future to move them out into a separate module and write tests but that's a bigger undertaking. Not sure it's worth adding another functional test: our general `lr` test has negative coefficients while `nnlr` test has all positive coefficients.

rsmtool/configuration_parser.py
```diff
@@ -954,8 +970,8 @@ def validate_config(cls, config, context='rsmtool'):
         # linear regression model
         if new_config['skll_objective']:
             if not is_skll_model(new_config['model']):
-                logging.warning("You specified a custom SKLL objective but also chose a "
-                                "non-SKLL model. The objective will be ignored.")
+                warnings.warn("You specified a custom SKLL objective but also chose a "
```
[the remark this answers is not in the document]
aloukina: Yes - we are already using them in many other places (e.g. in `utils.metrics`: https://github.com/EducationalTestingService/rsmtool/blob/master/rsmtool/utils/metrics.py). See also the discussion here for example: https://github.com/EducationalTestingService/rsmtool/issues/367 .
I replaced logging with warning in this module following the table you linked: keep `logging` for status monitoring, use `warning` when the issue is avoidable and the user should modify their code.

rsmtool/notebooks/builtin_model.ipynb
desilinguist: Sure, may be we should file an issue for adding such tests in the future?

rsmtool/configuration_parser.py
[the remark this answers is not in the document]
desilinguist: Makes total sense! Thanks for reminding me.

rsmtool/notebooks/builtin_model.ipynb
aloukina: https://github.com/EducationalTestingService/rsmtool/issues/470

desilinguist (approved):
Thanks!

mulhod (approved):

PR #477, opened 2020-10-15
changes the master branch to main and updates references to master with main in the codebase

srhrshr:


doc/index.rst
```diff
@@ -1,4 +1,4 @@
-.. RSMTool documentation master file, created by
+.. RSMTool documentation main file, created by
```
desilinguist, line 1: We don't really need to change it here since that's coming from a Sphinx template.

srhrshr: My bad! Reverted this change.

desilinguist (approved):
LGTM. @aloukina can you please also take a look?

aloukina (approved):
Looks good to me - thank you!

PR #484, opened 2020-10-29
Update report generation to work with latest nbconvert

### LINKED ISSUE #471: Unpin nbconvert when upstream fix is released
Currently we have pinned `nbconvert`  since v6.0 and above [break](https://github.com/jupyter/nbconvert/issues/1394) custom templates. However, there's a fix probably on the way at some point. In the next release, we should remove the pin assuming the fix has been released.

desilinguist:
This PR closes #471. 

- Update the reporting code and template.
- Remove nbconvert pin.
- Include asyncio + CPython 3.8  workaround to make sure things work on Windows.

### DIFF
--- a/rsmtool/reporter.py
+++ b/rsmtool/reporter.py
@@ -320,8 +320,9 @@ class Reporter:
         # set a high timeout for datasets with a large number of features
         report_config = Config({'ExecutePreprocessor': {'enabled': True,
                                                         'timeout': 3600},
-                                'HTMLExporter': {'template_path': [template_path],
-                                                 'template_file': 'report.tpl'}})
+                                'HTMLExporter': {"template_name": "classic",
+                                                 "template_file": join(template_path,
+                                                                       'report.tpl')}})
 
         exportHtml = HTMLExporter(config=report_config)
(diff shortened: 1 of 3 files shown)


mulhod (approved):
LGTM! I tested out environment creation and some tests that generate reports. I didn't try on Windows, though. Just wanted to give an extra check for OSX.

rsmtool/reporter.py
```diff
@@ -317,11 +319,21 @@ def convert_ipynb_to_html(notebook_file, html_file):
         :ref:`render_notebook <render_notebook>` command-line utility.
         """
 
+        # `nbconvert` uses `asyncio` which uses an entirely default
```
aloukina, line 322: Would Azure be sufficient for testing this? (Not that we can do much about it)

aloukina (approved):
Looks good to me - thank you for figuring this out!

rsmtool/reporter.py
desilinguist: Yes, the Azure build was failing without this which is how I got turned on to this issue.

PR #493, opened 2020-12-30
Pre commit

mulhod:
Introducing `pre-commit` into the workflow. I have tried to segregate my commits into those that add/change config files and those that apply changes to source files that had `flake8` issues, `isort` issues, etc.

One other big change here is moving the development dependencies into a separate file. This will allow the Conda/pip packages to be leaner with respect to dependencies and it would also enable things like:

```
pip install -e .[dev]
```

This would install `rsmtool` along with all of its dependencies and extra development dependencies.

At this point, this is just a draft PR. ~~I hope the build works, but I'm not confident it will.~~


mulhod:
If you want to review this, I highly recommend stepping through the commits. The number of changed files is very high.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,20 @@
...
+# See https://pre-commit.com/hooks.html for more git_hooks
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.3.0
+    hooks:
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-yaml
+        exclude: conda-recipe/rsmtool/meta.yaml
+    -   id: check-added-large-files
```
desilinguist, line 11: Looks like the default limit here is 500 kB. Does that cover all of our test data?

desilinguist: How about also adding `debug-statements`? I have done that in the past.

setup.cfg
```diff
@@ -0,0 +1,7 @@
+[flake8]
+ignore = E203, E266, E501, W503, F403, C0103, R0913, E1103, W0142, E0611, E1101, C0330, C0326, W504
+max-line-length = 88
+max-complexity = 18
+select = B, C, E, F, W, T4, B9
+exclude =
+    doc/conf.py
```
desilinguist, line 7: Can this not go in the toml file also?

setup.py
```diff
@@ -38,7 +37,10 @@ def requirements():
                      'render_notebook = rsmtool.reporter:main',
                      'convert_feature_json = rsmtool.convert_feature_json:main']
                     },
-      install_requires=requirements(),
+      install_requires=requirements('requirements.txt'),
+      extras_require={
+          'dev': requirements('dev-requirements.txt')
```
desilinguist, line 42: Nice!

desilinguist (changes requested):
This looks really nice @mulhod ! Thanks for putting this together. This should make things a lot better during development. I made some minor suggestions but I also have two overarching comments: 

1. We shouldn't apply the formatting changes to the SVG files in case something breaks.
2. Can you please double check the documentation by building from scratch to make sure there are no errors and nothing looks weird or breaks after you make these formatting changes to the RST files?

.pre-commit-config.yaml
mulhod: For each check, I ran the check on every single file, so the `check-added-large-files` check was run on all test data. We can increase this limit, though, if you want.

I will add in `debug-statements`.

setup.cfg
mulhod: Unfortunately, it appears `flake8` still doesn't support `pyproject.toml`: https://gitlab.com/pycqa/flake8/-/issues/428

.pre-commit-config.yaml
desilinguist: Ah, if that was already run then it's fine. We can tweak the limit in future PRs if and when we add files larger than the default limit.

mulhod:
I will add an exclusion for the svg files for things like adding EOF space, etc.

setup.cfg
desilinguist: Well that's a shame.

mulhod:
I'm checking on the documentation now. I had built it to at least see my changes and things looked fine. But I will check more pages that had changes.

Excluding svg files from the reformatting checks reduced the number of changed files quite a bit, so hopefully it's easier to review.

desilinguist:
Hmm, one other thing, is this isort configuration also picked up by the isort sublime plugin?

mulhod:
> Hmm, one other thing, is this isort configuration also picked up by the isort sublime plugin? And by the flake8 plugin as well?

Yes, according to the documentation I read about the plugin, it picks up the `pyproject.toml` file.

mulhod:
I went through a lot of changes in the documentation and could not find any cases that affected the result.

mulhod:
Consider not recommending the `isort` plugin OR getting rid of the `isort` hook (and potentially others).

doc/contributing.rst
```diff
@@ -33,6 +37,39 @@ Note that the file ``doc/requirements.txt`` is meant specifically for the ReadTh
...
+
+    pre-commit run
+
+To run the ``isort`` hook alone on changed files::
+
+    pre-commit run isort
+
+To run the ``isort`` hook alone on a given file::
+
+    pre-commit run isort <file-path>
```
mulhod, line 63:
```suggestion
    pre-commit run isort --files <file-path>
```

PR #505, opened 2021-03-01
502 remove support for xls files

### LINKED ISSUE #502: Remove support for xls files
Newer versions of `pandas` will be dropping support for `xls` files so we need to do the same.

aloukina:
This PR closes #502. 

* Removed support for xls files
* Added note to the documentation to highlight this


rsmtool/reader.py
```diff
@@ -239,7 +239,7 @@ def read_from_file(filename, converters=None, **kwargs):
         if file_extension in ['.csv', '.tsv']:
             sep = '\t' if file_extension == '.tsv' else ','
             do_read = partial(pd.read_csv, sep=sep, converters=converters)
-        elif file_extension in ['.xls', '.xlsx']:
+        elif file_extension == '.xlsx':
             do_read = partial(pd.read_excel, converters=converters)
         elif file_extension in ['.sas7bdat']:
```
desilinguist, line 244:
```suggestion
        elif file_extension == '.sas7bdat':
```

Just to be consistent?

tests/test_reader.py
```diff
@@ -157,7 +157,7 @@ def test_container_train_property(self):
     def test_container_feature_specs_property(self):
         test_lists = [[('feature_specs', 'csv'), ('test', 'tsv')],
                       [('train', 'csv'), ('feature_specs', 'xlsx')],
-                      [('train', 'csv'), ('feature_specs', 'xls'), ('train_metadata', 'tsv')],
+                      [('train', 'csv'), ('feature_specs', 'tsv'), ('train_metadata', 'tsv')],
```
desilinguist, line 160: Why "tsv" instead of "xlsx"?

doc/pipeline.rst
```diff
@@ -24,10 +24,14 @@ Input file format
...
 
-RSMTool supports input files in ``.csv``, ``.tsv``, ``.sas7bdat``, ``xls``/``.xlsx``, or ``.jsonlines`` format. The format of the file is determined based on the extension. In all cases the :ref:`output files<intermediate_files_rsmtool>` will be saved in ``.csv`` format by default (see :ref:`file format <file_format>` for other intermediate file output options).
+RSMTool supports input files in ``.csv``, ``.tsv``, ``.sas7bdat``, ``.xlsx``, or ``.jsonlines`` format. The format of the file is determined based on the extension. In all cases the :ref:`output files<intermediate_files_rsmtool>` will be saved in ``.csv`` format by default (see :ref:`file format <file_format>` for other intermediate file output options).
 
 For ``Excel`` spreadsheets, all data must be stored in the first sheet. 
 
+.. note::
+
+    RSMTool 8.1 and higher no longer support ``.xls`` files. All Excel files must be in ``.xlsx`` format.
```
desilinguist, line 33: Do we want to say why?

desilinguist (changes requested):
Looks great! Just some minor comments/suggestions. Also I see that you added `lr_rsmtool_rsmpredict.model` file? Was that an accident?

doc/pipeline.rst
aloukina: I thought of it but decided against it: the format is very old by now so I don't think we need to further justify why we no longer support it. We can add something if you think it's important.

desilinguist: Fair enough. No ,I think it's fine.

tests/test_reader.py
aloukina: Because we are testing the combination `[('train', 'csv'), ('feature_specs', 'xlsx')]` in the previous test so I wanted to have a different combination of file type + file format.

desilinguist: Got it!

aloukina:
I removed the .model file: I am puzzled why it was added...

desilinguist (approved):

mulhod (approved):
LGTM!

PR #513, opened 2021-05-05
Fix broken RSMTool CI build

Frost45:
Made changes to suppress all warnings when not in STRICT mode instead of just deprecation warnings.

### DIFF
--- a/rsmtool/test_utils.py
+++ b/rsmtool/test_utils.py
@@ -40,7 +40,8 @@ tools_with_output = ['rsmtool', 'rsmeval',
 # if so, any deprecation warnings found in HTML
 # reports should not be ignored
 STRICT_MODE = os.environ.get('STRICT', None)
-IGNORE_DEPRECATION_WARNINGS = False if STRICT_MODE else True
+# IGNORE_DEPRECATION_WARNINGS = False if STRICT_MODE else True
+IGNORE_WARNINGS = False if STRICT_MODE else True
 
 
 def check_run_experiment(source,
@@ -143,8 +144,10 @@ def check_run_experiment(source,
     # make sure that there are no warnings in the report
     # but ignore deprecation warnings if appropriate
     warning_msgs = collect_warning_messages_from_report(html_report)
-    if IGNORE_DEPRECATION_WARNINGS:
-        warning_msgs = [msg for msg in warning_msgs if 'DeprecationWarning' not in msg]
+    # if IGNORE_DEPRECATION_WARNINGS:
+    #     warning_msgs = [msg for msg in warning_msgs if 'DeprecationWarning' not in msg]
+    if IGNORE_WARNINGS:
+        warning_msgs = [msg for msg in warning_msgs if 'Warning' not in msg]
     assert_equal(len(warning_msgs), 0)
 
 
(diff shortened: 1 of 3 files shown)


rsmtool/test_utils.py
```diff
@@ -790,20 +790,26 @@ def check_report(html_file,
     report_errors = 0
     report_warnings = 0
 
+    # Setting raise_warnings to false if not in STRICT mode
+    if IGNORE_WARNINGS:
+        raise_warnings = False
+
     with open(html_file, 'r') as htmlf:
-        for line in htmlf:
+        for i, line in enumerate(htmlf):
```
desilinguist, line 798: What is the enumerate for? I don't see `i` being used anywhere?

desilinguist (approved):
This looks good @Frost45! One minor comment and a question: you had the same errors in `main` as `stable` before this fix and they were all resolved once this fix was applied?

rsmtool/test_utils.py
```diff
@@ -141,10 +141,10 @@ def check_run_experiment(source,
     check_report(html_report, raise_warnings=False)
 
     # make sure that there are no warnings in the report
-    # but ignore deprecation warnings if appropriate
+    # but ignore warnings if appropriate
     warning_msgs = collect_warning_messages_from_report(html_report)
-    if IGNORE_DEPRECATION_WARNINGS:
-        warning_msgs = [msg for msg in warning_msgs if 'DeprecationWarning' not in msg]
+    if IGNORE_WARNINGS:
+        warning_msgs = [msg for msg in warning_msgs if 'Warning' not in msg]
```
mulhod, line 147: Just wondering, won't "Warning" always be in the message? Which messages would not be filtered out here?

rsmtool/test_utils.py
Frost45: @desilinguist Yeah, there were 70 errors or so, when not in STRICT mode and 90 in STRICT mode (With deprecation warnings). After this fix, running the tests normally does not throw any errors.

rsmtool/test_utils.py
mulhod: Same for the repetitions below.

rsmtool/test_utils.py
Frost45: Oh, I was trying out something but ended up not needing it. Should I revert this line?

desilinguist: Yes, please go ahead.

mulhod (approved):
Looks good! Just a question.

rsmtool/test_utils.py
desilinguist: Good question! I don't think we need this filtering since previously it was to filter out non-deprecation warnings :)

aloukina (changes requested):
We also need to update documentation here: https://rsmtool.readthedocs.io/en/stable/internal/release_process.html?highlight=warnings#rsmtool-release-process to remove "deprecation"

rsmtool/test_utils.py
Frost45: @desilinguist @mulhod I was actually considering not calling `collect_warning_messages_from_report()` when not in STRICT mode but did not want to change it too much. There was some issue with blank spaces, but I've fixed that. What do you guys think?

Frost45:
```suggestion
            if not IGNORE_WARNINGS:
                  warning_msgs = collect_warning_messages_from_report(html_report)
                  assert_equal(len(warning_msgs), 0)
```

mulhod: @Frost45 Looks good to me!

mulhod: Please make the change and let the build run.

rsmtool/test_utils.py
```diff
@@ -231,10 +231,10 @@ def check_run_evaluation(source,
     check_report(html_report, raise_warnings=False)
 
     # make sure that there are no warnings in the report
-    # but ignore deprecation warnings if appropriate
+    # but ignore warnings if appropriate
     warning_msgs = collect_warning_messages_from_report(html_report)
-    if IGNORE_DEPRECATION_WARNINGS:
```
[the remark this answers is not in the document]
Frost45: @aloukina I've made some changes to this code. Do you still think it should be made into a function?

desilinguist (approved):
Latest changes look great!!! 🎉

mulhod (approved):
LGTM!

rsmtool/test_utils.py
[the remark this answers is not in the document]
aloukina: There is some repetition in the code, but I think it's fine to leave it as is.

aloukina (approved):
Thank you for doing this, looks great!

PR #521, opened 2021-05-25
Fixed Regex in parse_json_with_comments

### LINKED ISSUE #492: JSON loading fails on valid JSON with parse_json_with_comments
I ran into an issue recently where I was using `parse_json_with_comments` and it failed despite there not being comments, actually.

This issue is not really about `rsmtool` for the most part, but the issue I ran into could have easily been an issue for `rsmtool` configuration files (perhaps in the free-text description fields, for instance).

Here is the offending JSON file:
```
{
    "advisories": {},
    "metadata": {
        "program": null,
        "score_max": 5,
        "score_min": 1
    },
    "multiple_models": false,
    "predict": {
        "trim_max": null,
        "trim_min": null
    },
    "read_data": {
        "data_to_plaintext": [
            "textpipes.pipes.CharFixerDefault",
            "textpipes.pipes.ReplaceTabsWithSpace",
            "textpipes.pipes.MultiSpaceToSingleSpace",
            "textpipes.pipes.Strip"
        ]
    },
    "text_to_instance": null,
    "train": {
        "notes": "The training process was recorded here: https://etsorg1-my.sharepoint.com/personal/briordan_ets_org/_layouts/15/Doc.aspx?sourcedoc={f3190115-91e2-4aac-bf28-3556a1669636}&action=edit&wd=target%28predict-scores-operational.one%7C3642acf5-3099-db40-9ddd-dc564ccc8dfb%2Fiteration-3%7C6b4bfc6b-5d8a-9e43-a382-9cc84884662b%2F%29",
        "repository_commit": "12.2020_model_deployment_1.0",
        "repository_name": "strides-scoring-pt",
        "repository_url": "https://stash.research.ets.org:7994/projects/BR/repos/strides-scoring-pt/browse",
        "train_cmd": "futil.run_experiment --config /home/research/mmulholland/text-dynamic/strides-scoring-pt/strides_scoring_pt/training_config/expt-3/expt-3_run_config.jsonnet --config_changes /home/research/mmulholland/text-dynamic/strides-scoring-pt/strides_scoring_pt/training_config/expt-3/expt-3_all_config_changes.jsonnet"
    }
}
```

It was the `"notes"` field that caused problems. Anyway, I confirmed that I could actually load in the JSON just fine with regular-old `json.loads` (like I said, there are no comments actually). But I was looking around for other libraries that did comment removal from JSON and stumbled on [`jsoncomment`](https://pypi.org/project/jsoncomment/0.2.3/). This library was actually able to load the JSON file above.

If you think it's worthwhile, I could try it out here.

Frost45:
Regex to remove comments from JSON file in parse_json_with_comments was incorrectly matching URLs like "https://...". 

Fixed by adding negative lookbehind to regex.

Added additional unit tests for parse_json_with_comments

Closes #492.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -66,24 +66,98 @@ def test_convert_to_float():
 
 def test_parse_json_with_comments():
 
-    # Need to add comments
-    json_with_comments = ("""{"key1": "value1", /*some comments */\n"""
-                          """/*more comments */\n"""
-                          """"key2": "value2", "key3": 5}""")
+    json_with_comments = """
+    {{
+        "key1": "{0}", {1}
+        {1}
+        "key2": "value2", 
+        "key3": 5
+    }}"""
 
-    tempf = tempfile.NamedTemporaryFile(mode='w', suffix='.csv', delete=False)
+    # ensure that URLs are not filtered out
+    for value in ['value1', 'http://stash.research.ets.org:8000', 'https://stash.research.ets.org:8000']:
+        for comment in ["/*some comments */", "//some comments "]:
+            test_json_string = json_with_comments.format(value, comment)
+            expected_result = {'key1': value, 'key2': 'value2', 'key3': 5}
+            yield check_parse_json_with_comments, test_json_string, expected_result
+
+
+def test_parse_json_with_comments_no_comments():
+
+    test_json_string = r"""
+    {
+        "advisories": {},
+        "metadata": {
+            "program": null,
+            "score_max": 5,
+            "score_min": 1
+        },
+        "multiple_models": false,
+        "predict": {
+            "trim_max": null,
+            "trim_min": null
+        },
+        "read_data": {
+            "data_to_plaintext": [
+                "textpipes.pipes.CharFixerDefault",
+                "textpipes.pipes.ReplaceTabsWithSpace",
+                "textpipes.pipes.MultiSpaceToSingleSpace",
+                "textpipes.pipes.Strip"
+            ]
+        },
+        "text_to_instance": null,
+        "train": {
+            "notes": "The training process was recorded here: https://etsorg1-my.sharepoint.com/personal/briordan_ets_org/_layouts/15/Doc.aspx?sourcedoc={f3190115-91e2-4aac-bf28-3556a1669636}&action=edit&wd=target%28predict-scores-operational.one%7C3642acf5-3099-db40-9ddd-dc564ccc8dfb%2Fiteration-3%7C6b4bfc6b-5d8a-9e43-a382-9cc84884662b%2F%29",
+            "repository_commit": "12.2020_model_deployment_1.0",
+            "repository_name": "strides-scoring-pt",
+            "repository_url": "https://stash.research.ets.org:7994/projects/BR/repos/strides-scoring-pt/browse",
+            "train_cmd": "futil.run_experiment --config /home/research/mmulholland/text-dynamic/strides-scoring-pt/strides_scoring_pt/training_config/expt-3/expt-3_run_config.jsonnet --config_changes /home/research/mmulholland/text-dynamic/strides-scoring-pt/strides_scoring_pt/training_config/expt-3/expt-3_all_config_changes.jsonnet"
+        }
+    }"""
+
+    expected_result = {
+        "advisories": {},
+        "metadata": {
+            "program": None,
+            "score_max": 5,
+            "score_min": 1
+        },
+        "multiple_models": False,
+        "predict": {
+            "trim_max": None,
+            "trim_min": None
+        },
+        "read_data": {
+            "data_to_plaintext": [
+                "textpipes.pipes.CharFixerDefault",
+                "textpipes.pipes.ReplaceTabsWithSpace",
+                "textpipes.pipes.MultiSpaceToSingleSpace",
+                "textpipes.pipes.Strip"
+            ]
+        },
+        "text_to_instance": None,
+        "train": {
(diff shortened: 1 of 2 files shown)


mulhod:
@Frost45 It might be a good idea to include a unit test here.

desilinguist:
Yeah, we definitely need a new test here.

tests/test_utils.py
```diff
@@ -86,6 +86,116 @@ def test_parse_json_with_comments():
...
+def test_parse_json_with_comments_single_line_comments():
+
+    # Need to add comments
+    json_with_comments = ("""{"key1": "value1", //some comments \n"""
+                          """//more comments \n"""
+                          """"key2": "value2", "key3": 5}""")
+
+    tempf = tempfile.NamedTemporaryFile(mode='w', suffix='.csv', delete=False)
+    filename = tempf.name
+    tempf.close()
```
mulhod, line 98: I could be wrong but I think you can just use `tempf` in place of `buff` below, can't you? That is, I think you can get rid of this line and the `with` part below.

```
❯ ~/opt/python/3.8/bin/ipython
import tempfile
Python 3.8.1 | packaged by conda-forge | (default, Jan 29 2020, 15:06:10)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.12.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import tempfile

In [2]: json_with_comments = ("""{"key1": "value1", //some comments \n"""
   ...:                           """//more comments \n"""
   ...:                           """"key2": "value2", "key3": 5}""")

In [3]: tempf = tempfile.NamedTemporaryFile(mode='w', suffix='.csv', delete=False)

In [4]: filename = tempf.name

In [5]: tempf.write(json_with_comments)
Out[5]: 81

In [6]: tempf.close()

In [7]: filename
Out[7]: '/var/folders/f_/_561m4hx5cz3tn5hj63ryvjxvw9dhw/T/tmpc4cithhk.csv'

In [8]: ! cat /var/folders/f_/_561m4hx5cz3tn5hj63ryvjxvw9dhw/T/tmpc4cithhk.csv
{"key1": "value1", //some comments
//more comments
"key2": "value2", "key3": 5}
```

tests/test_utils.py
```diff
@@ -86,6 +86,116 @@ def test_parse_json_with_comments():
     eq_(result, {'key1': 'value1', 'key2': 'value2', 'key3': 5})
 
 
+def test_parse_json_with_comments_single_line_comments():
+
+    # Need to add comments
+    json_with_comments = ("""{"key1": "value1", //some comments \n"""
+                          """//more comments \n"""
+                          """"key2": "value2", "key3": 5}""")
```
mulhod, line 94: I feel like there's a way to make this more readable. Not sure this is it, though.

```suggestion
    json_with_comments = """
    {
        "key1": "value1", //some comments
        //more comments
        "key2": "value2", "key3": 5
    }"""
```

mulhod (commented):
Minor suggestions. Looks good!

tests/test_utils.py
Frost45: Sounds good! I'll make that change!

tests/test_utils.py
```diff
@@ -66,17 +66,41 @@ def test_convert_to_float():
...
-                          """"key2": "value2", "key3": 5}""")
+    json_with_comments = """
+    {
+        "key1": "value1", /*some comments */
+        /*more comments */
+        "key2": "value2", 
+        "key3": 5
+    }"""
 
     tempf = tempfile.NamedTemporaryFile(mode='w', suffix='.csv', delete=False)
```
desilinguist, line 74: I wonder why we chose to use "csv" as the extension?

desilinguist (commented):
@mulhod since you originally reported this issue, can you test this function with the actual file you were trying to use and make sure it works?

tests/test_utils.py
```diff
@@ -66,17 +66,41 @@ def test_convert_to_float():
 
 def test_parse_json_with_comments():
```
mulhod, line 67: One more thing I was thinking: This test is pretty much the same test with slight modifications. Could the test simply be parameterized or could there be one test that uses a helper function instead? Even just a list of values with a for loop (or two for loops) might be good, e.g.

```python
for json_string in [json_string_inline, json_string_newline]:
    for http_or_https in ["http", "https"]:
        json_string = ....format(http_or_https)
        [do the rest of the stuff] 
```

desilinguist: Good idea!

mulhod:
@desilinguist Yeah, I will try it out. That file is actually in the issue connected with this PR.

desilinguist:
> @desilinguist Yeah, I will try it out. That file is actually in the issue connected with this PR.

Ah then we should turn that into a test too @Frost45!

tests/test_utils.py
Frost45: @mulhod @desilinguist I thought that different functionalities or test cases must be tested in separate functions? Is that not the case? Otherwise, these test functions can be modified to test everything in one function, with two for loops.

Frost45:
> > @desilinguist Yeah, I will try it out. That file is actually in the issue connected with this PR.
> 
> Ah then we should turn that into a test too @Frost45!

Sure, I've tested it manually but I can add it to a test function too.

tests/test_utils.py
mulhod: I don't think we need to have separate tests. If we use a test generator we can kind of get the best of both worlds, perhaps.

desilinguist: Yeah, using a test generator is what I was thinking. That way you get the code reduction but also each test being logically separate.

mulhod: https://nose.readthedocs.io/en/latest/writing_tests.html#test-generators

Check this out.

tests/test_utils.py
Frost45: Should I change it to .json?

mulhod: Yeah, I think it would be best.

desilinguist: Agreed.

desilinguist (approved):
Looks great now!

mulhod (approved):
LGTM 🎉

PR #525, opened 2021-06-02
Update test files for linear svr

Frost45:
Updated input and output test files for linear svr.

These files can be unstable and change due to changes in SKLL or sklearn. However, as they are inputs for testing rsmcompare and rsmsummarize, they need to be updated.


desilinguist (approved):
Finally! 😮‍💨

Issue #534, opened 2021-07-09
Tests which involve jsonlines no longer pass


aloukina:
After the recent update (July 6, 2021) tests involving jsonlines no longer pass:

- Tests in `test_reader.TestJsonLines`

- Test in test_reader.TestDataReader when input is in `jsonlines` format

And the related regression tests:

- `test_experiment_rsmtool_2.test_run_experiment_parameterized('lr-nested-jsonlines-input-files', 'lr_nested_jsonlines_input_files', {})`

- `test_experiment_rsmtool_2.test_run_experiment_parameterized('lr-jsonlines-input-files', 'lr_jsonlines_input_files', {})`

- `test_experiment_rsmpredict.test_run_experiment_parameterized('lr-predict-nested-jsonlines-input-files', {})`

- `test_experiment_rsmpredict.test_run_experiment_parameterized('lr-predict-jsonlines-input-files', {})`

- `test_experiment_rsmeval.test_run_experiment_parameterized('lr-eval-nested-jsonlines-input-files', 'lr_eval_nested_jsonlines_input_files', {})`

- ` test_experiment_rsmeval.test_run_experiment_parameterized('lr-eval-jsonlines-input-files', 'lr_eval_jsonlines_input_files', {})`

Issue #533, opened 2021-07-09
Errors in TestAnalyzer


aloukina:
TestAnalyzer generates multiple errors after recent update, mostly related to correlations. The regression tests are fine.

Issue #531, opened 2021-07-09
Pandas FutureWarnings


aloukina:
We get the following FutureWarnings in Pandas that need to be fixed throughout the code: 

`FutureWarning: In a future version of pandas all arguments of DataFrame.drop except for the argument 'labels' will be keyword-only [multiple occurrences]`

We also get another warning but it is coming from Statmodels:

```
$HOME/miniconda3/envs/rsmdev/lib/python3.9/site-packages/statsmodels/tsa/tsatools.py:142: FutureWarning: In a future version of pandas all arguments of concat except for the argument 'objs' will be keyword-only
  x = pd.concat(x[::order], 1)
FutureWarning: In a future version of pandas all arguments of concat except for the argument 'objs' will be keyword-only
```

PR #535, opened 2021-07-09
Fix issues related to new Pandas version

aloukina:
This PR addresses #534 , #533  , and #531 

* Fixed test set up in TestAnalyzer
* Changed read_jsonlines to accommodate new behavior of `pd.normalize_json`
* Added "axis" keyword to instances of `pandas.dataframe.drop`

### DIFF
--- a/rsmtool/reader.py
+++ b/rsmtool/reader.py
@@ -68,12 +68,12 @@ def read_jsonlines(filename, converters=None):
             df_column = pd.json_normalize(df[column])
 
             # Starting with Pandas v1.3, we get an empty data frame
-            # if the column does not contain a nested json. 
+            # if the column does not contain a nested json.
             # If this is the case, we simply copy the column.
             if df_column.empty:
                 df_column = df[column].copy()
 
-        # Pandas <v1.3 will raise an attribute error instead, 
+        # Pandas <v1.3 will raise an attribute error instead,
         # so we'll catch that too
         except AttributeError:
             df_column = df[column].copy()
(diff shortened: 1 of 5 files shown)


aloukina:
I kept the error catching in `read_jsonlines` for compatibility with earlier Pandas version: that line won't be covered by tests.

rsmtool/reader.py
```diff
@@ -63,8 +63,18 @@ def read_jsonlines(filename, converters=None):
 
     dfs = []
     for column in df:
+        # let's try to normalize this column
         try:
             df_column = pd.json_normalize(df[column])
+
+            # from Pandas 1.3 we get an empty df
+            # if the column does not contain a nested json
+            # in this case we simply copy the column
```
desilinguist, line 72:
```suggestion
            # Starting with Pandas v1.3, we get an empty data frame
            # if the column does not contain a nested json. 
            # If this is the case, we simply copy the column.
```

rsmtool/reader.py
```diff
@@ -63,8 +63,18 @@ def read_jsonlines(filename, converters=None):
...
+            if df_column.empty:
+                df_column = df[column].copy()
+
+        # Pandas <1.3 will raise an attribute error:
+        # we'll catch that too
```
desilinguist, line 77:
```suggestion
        # Pandas <v1.3 will raise an attribute error instead, 
        # so we'll catch that too
```

desilinguist (changes requested):
Looks great! Just some minor tweaks to the code comments.

aloukina:
Thank you, @desilinguist  - all fixed!

desilinguist (approved):

PR #562, opened 2022-03-01
Fix `use_thumbnails` option.

### LINKED ISSUE #553: `use_thumbnails` not working correctly
It seems that setting `use_thumbnails` to true with the latest versions of matplotlib causes _both_ the thumbnail and the full figure to show in the generated report. This is because the `savefig()` call we make before showing the thumbnail both saves the figures as well as displays it. 

The solution should be simply to add `plt.ioff()` in all the header notebook(s) to simply disable interactive plotting.

desilinguist:
This PR resolves #553. 

Adding `use_thumbnails = true` to a configuration file was generating both the thumbnail as well as the full-sized figure. The solution was to turn off interactive plotting in all header notebooks. 

To review, please run 2 `rsmxval` experiments:  one that uses `use_thumbnails`  and one that doesn't. Make sure that the figures show up as expected in all the reports in both experiments. Please also try an `rsmcompare` experiment separately since that is not part of `rsmxval`.

### DIFF
--- a/rsmtool/notebooks/comparison/header.ipynb
+++ b/rsmtool/notebooks/comparison/header.ipynb
@@ -47,6 +47,7 @@
     "\n",
     "from IPython import sys_info\n",
     "from IPython.display import display, HTML, Image, Javascript, Markdown, SVG\n",
+    "from matplotlib import pyplot as plt\n",
     "\n",
     "from rsmtool.comparer import Comparer\n",
     "\n",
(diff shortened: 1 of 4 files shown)


bndgyawali (approved):
I ran with and without "use_thumbnails" and I see "use_thumbnails" displays thumbnail images while without it the larger images.

Frost45 (approved):
Tested it and it works great! 🎉

PR #565, opened 2022-03-04
Make feature boxplots by subgroups more flexible.

### LINKED ISSUE #560: Increase feature limit for subgroup boxplots
Currently, we do not display feature subgroup boxplots for > 30 features. However, this may be too conservative in some cases. We should increase this limit to 100.

desilinguist:
This PR resolves #560. 

- Change the limit when no boxplots are shown to 150 and make it easier to see message when boxplots are omitted by adding a heading.

- Add a new message about using thumbnails for features > 30 but <= 150, if not specified.

To test this, I’d recommend generating pandas dataframes with 50 and 200 columns containing random values and then adding an ID,  score, and a subgroup column and then running RSMTool experiments.


desilinguist:
D’oh! Will fix and then you can quickly check.

mulhod:
Taking a long time to generate figures, etc. for 200+ columns!

mulhod:
It did eventually do it though.

mulhod (approved):

Frost45 (approved):
Generated new dataframes and I ran into the same error as Matt. After pulling the latest version, it all works!

PR #581, opened 2023-01-05
Add new `fast_predict()` API method for prediction

### LINKED ISSUE #546: Refactor compute_and_save predictions
We want to refactor `compute_and_save_predictions` into three separate steps:

(a) Read in model parameters from file into memory.
(b) Apply preprocessing, generate predictions and post-process predictions using model parameters stored in memory and return a df with predictions.
(c) Save predictions to disc.

This will allow us to generate predictions in production without writing and reading from disc for each response.

desilinguist:
Closes #546. 

The current method that `rsmpredict` uses is called `compute_and_save_prediction()` and is meant for batch prediction – it reads files from disks, does a lot of validation, and writes the predictions to disk. Therefore, it is not suited for real-time prediction on single instances which needs to happen in memory. 

This PR adds a new function called `fast_predict()` to `rsmpredict.py` meant to run on a single data point rather than a batch. This function expects the user to have already read all of the necessary files from disk (e.g., in a RFS service's `setup()` function) and simply pass the in-memory data structures to this function. This function then pre-processes the features, generates predictions for the single data point using the model, and then post-processes the predictions. 

In addition to adding this new function, this PR also adds tests for it and updates the API documentation to include this new function.

### DIFF
--- a/rsmtool/rsmpredict.py
+++ b/rsmtool/rsmpredict.py
@@ -72,7 +72,7 @@ def fast_predict(
         existing model file in the "output" directory of the previously run
         RSMTool experiment.
     df_feature_info : pandas DataFrame
-        A pandas dataframe containing the information regarding the model features.
+        A DataFrame containing the information regarding the model features.
         The index of the dataframe should be the names of the features and
         the columns should be:
 
@@ -163,25 +163,21 @@ def fast_predict(
     # now compute the raw prediction for the given features
     df_predictions = modeler.predict(df_processed_features)
 
-    # trim raw predictions if ``trim_min`` and ``trim_max`` were specified
-    if trim_min and trim_max:
-        df_predictions["raw_trim"] = preprocessor.trim(
-            df_predictions["raw"], trim_min, trim_max, trim_tolerance
-        )
-        df_predictions["raw_trim_round"] = np.rint(df_predictions['raw_trim']).astype("int64")
-
-    # compute scaled predictions if the scaling parameters are available
+    # compute scaled prediction if the scaling parameters are available
     if train_predictions_mean and train_predictions_sd and h1_mean and h1_sd:
         df_predictions["scale"] = (
             (df_predictions["raw"] - train_predictions_mean) / train_predictions_sd
         ) * h1_sd + h1_mean
 
-        # apply trimming to scaled scores too, if possible
-        if trim_min and trim_max:
-            df_predictions["scale_trim"] = preprocessor.trim(
-                df_predictions["scale"], trim_min, trim_max, trim_tolerance
+    # trim both raw and scale predictions if ``trim_min`` and ``trim_max`` were specified
+    if trim_min and trim_max:
+        for column in df_predictions.columns.drop("spkitemid"):
+            df_predictions[f"{column}_trim"] = preprocessor.trim(
+                df_predictions[column], trim_min, trim_max, trim_tolerance
             )
-            df_predictions["scale_trim_round"] = np.rint(df_predictions['scale_trim']).astype("int64")
+            df_predictions[f"{column}_trim_round"] = np.rint(
+                df_predictions[f"{column}_trim"]
+            ).astype("int64")
 
     # return the predictions as a dictionary but without the ID column
     return df_predictions.drop("spkitemid", axis="columns").to_dict(orient="records")[0]
--- a/rsmtool/rsmpredict.py
+++ b/rsmtool/rsmpredict.py
@@ -48,7 +48,7 @@ def fast_predict(
     as much as possible.
 
     This function should only be used when the goal is to generate predictions
-    using RSMTool models in production. The user should everything from disk
+    using RSMTool models in production. The user should read everything from disk
     in a separate thread/function and pass the inputs to this function.
 
     Parameters
@@ -94,10 +94,10 @@ def fast_predict(
 
     Returns
     -------
-    pandas DataFrame
-        A dataframe containing the raw, scaled, trimmed, and rounded
+    dict[str, float]
+        A dictionary containing the raw, scaled, trimmed, and rounded
         predictions for the input features. It contains the following
-        columns: "raw", "scale", "raw_trim", "scale_trim", "raw_trim_round",
+        keys: "raw", "scale", "raw_trim", "scale_trim", "raw_trim_round",
         and "scale_trim_round".
 
     Raises
@@ -169,8 +169,8 @@ def fast_predict(
         trim_tolerance,
     )
 
-    # return the predictions without the ID column
-    return df_predictions.drop("spkitemid", axis="columns")
+    # return the predictions as a dictionary but without the ID column
+    return df_predictions.drop("spkitemid", axis="columns").to_dict(orient="records")[0]
 
 
 def compute_and_save_predictions(config_file_or_obj_or_dict,
--- a/rsmtool/rsmpredict.py
+++ b/rsmtool/rsmpredict.py
@@ -32,8 +32,13 @@ def fast_predict(
     input_features,
     modeler,
     df_feature_info,
-    df_postprocessing_params,
-    predict_expected=False,
+    trim_min=None,
+    trim_max=None,
+    trim_tolerance=0.4998,
+    train_predictions_mean=None,
+    train_predictions_sd=None,
+    h1_mean=None,
+    h1_sd=None,
     logger=None,
 ):
     """
@@ -50,6 +55,9 @@ def fast_predict(
     This function should only be used when the goal is to generate predictions
     using RSMTool models in production. The user should read everything from disk
     in a separate thread/function and pass the inputs to this function.
+
+    Note that this function only computes regular predictions, not expected
+    scores.
 
     Parameters
     ----------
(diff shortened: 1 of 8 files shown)


rsmtool/rsmpredict.py
```diff
@@ -25,6 +28,151 @@
...
+    The main difference between this function and the ``compute_and_save_predictions()``
+    function is that the former is meant for batch prediction and reads
+    all its inputs from disk and writes its outputs to disk. This function,
+    however, is meant for real-time inference rather than batch. To this end,
+    it operates *entirely* in memory. Note that there is still a bit of overlap
+    between the two computation paths since we want to use the RSMTool API
+    as much as possible.
+
+    This function should only be used when the goal is to generate predictions
+    using RSMTool models in production. The user should everything from disk
```
mulhod, line 51:
```suggestion
    using RSMTool models in production. The user should read everything from disk
```

rsmtool/rsmpredict.py
```diff
@@ -25,6 +28,151 @@
...
+        Predict expected scores for classifiers that return probability
+        distributions over score. Note also that this assumes that the score
+        range consists of contiguous integers. Defaults to ``False``.
+    logger : logging object, optional
+        A logging object. If ``None`` is passed, get logger from ``__name__``.
+        Defaults to ``None``.
+
+    Returns
+    -------
+    pandas DataFrame
```
mulhod, line 97: It's not a very big deal, but could we just return a dictionary since it's really just one `Series` we're getting?

mulhod (approved):
This is great. Definitely been looking for something like this for awhile! (I still like the idea of being able to save something like these steps directly in a model file, but I think this is at least going in a good direction.)

I have a couple minor comments. One is just about the return value of the new function.

mulhod (approved):

rsmtool/rsmpredict.py
```diff
@@ -25,6 +28,151 @@
...
+        df_processed_features, _ = preprocessor.preprocess_new_data(
+            df_input_features, df_feature_info
+        )
+    except ValueError:
+        raise ValueError("Input features must not contain non-numeric values.") from None
+
+    # read the post-processing parameters
+    trim_min = df_postprocessing_params["trim_min"].values[0]
+    trim_max = df_postprocessing_params["trim_max"].values[0]
+    h1_mean = df_postprocessing_params["h1_mean"].values[0]
```
amandameganchan, line 129: is it possible to make passing these means and std devs optional in case we don't want/need scaled scores?

mulhod: Oh, yeah, good point!

desilinguist: You always need the `trim_min` and `trim_max` since that's needed to trim the raw predictions. So, you will need to pass in the post processing data frame anyway.

amandameganchan: I was just thinking of a case where we can define the trim_min and trim_max outside of this data frame.. the couple instances I have seen of using RSMTool models in the engine have this information in the model config or defined in the service itself

desilinguist: I see. Okay, how about this?

1. Make `trim_min`, `trim_max`, `trim_tolerance`, `train_predictions_mean`, train_predictions_sd`, `h1_mean` and `h1_sd` _all_ optional arguments for this function. All of them will be `None`, except for `trim_tolerance` which will have a default value of 0.4998. 
2. If you don't specify any of them, you just get the "raw" prediction.
3. If you specify `trim_min` and `trim_max`, you also get "raw_trim" and "raw_trim_round".
4. If you specify the other 4, you also get "scale", "scale_trim", and "scale_trim_round".

amandameganchan: that sounds great!

rsmtool/rsmpredict.py
```diff
@@ -25,6 +28,151 @@
...
+def fast_predict(
+    input_features,
+    modeler,
+    df_feature_info,
+    df_postprocessing_params,
+    predict_expected=False,
+    logger=None,
+):
+    """
+    Compute predictions for a single instance against given model.
```
tazin-afrin, line 40: Can this function be open to multiple instances?

desilinguist: What would be the use case? The primary use for this function is in an engine scoring service which would only operate on a single instance at a time.

tazin-afrin: We dont want any real-time prediction on batch instances? Sorry, I did not use rsmpredict much, so maybe my concepts are not that clear.

desilinguist: Batch prediction is not really used in prediction. That's mostly only used in experimentation and model-building mode.

desilinguist:
@mulhod, @amandameganchan, @tazin-afrin  I have made the requested changes. Please re-review.

tests/test_experiment_rsmpredict.py
```diff
@@ -320,3 +325,121 @@ def test_run_experiment_predict_expected_scores_non_probablistic_svc():
...
+                                   "lr-predict",
+                                   "existing_experiment",
+                                   "output")
+    feature_info_file = join(existing_experiment_dir, "lr_feature.csv")
+    model_file = join(existing_experiment_dir, "lr.model")
+
+    # read in the files
+    df_feature_info = pd.read_csv(feature_info_file, index_col=0)
+
+    # initialize the modelr instance
```
blongwill, line 431: Typo: modeler

tests/test_experiment_rsmpredict.py
```diff
@@ -320,3 +325,121 @@ def test_run_experiment_predict_expected_scores_non_probablistic_svc():
                        source,
                        'rsmpredict.json')
     do_run_prediction(source, config_file)
+
+
+def check_fast_predict(source, do_trim=False, do_scale=False):
+    """Ensure that predictions from `fast_predict()` match expected preditions."""
```
blongwill, line 331: Typo — expected  predictions.

rsmtool/rsmpredict.py
```diff
@@ -25,6 +28,165 @@
...
+        generate the model predictions. The keys should be names of the
+        features on which the model was trained and the values should
+        be the *raw* feature values.
+    modeler : rsmtool.modeler.Modeler object
+        The RSMTool modeler object using which the predictions
+        are to be generated. This object should be created from the already
+        existing model file in the "output" directory of the previously run
+        RSMTool experiment.
+    df_feature_info : pandas DataFrame
+        A pandas dataframe containing the information regarding the model features.
```
blongwill, line 75: Here you have DataFrame all lowercase even though you’re referring to the official object name. In a different one you have ‘Data frame’ as if it’s referring to a general frame of data. Not sure if this is intentional but maybe we should refer to it the same way for all of them to keep it uniform. I guess there’s no official convention so maybe I’m just splitting hairs.

rsmtool/rsmpredict.py
```diff
@@ -25,6 +28,165 @@
...
+
+        # apply trimming to scaled scores too, if possible
+        if trim_min and trim_max:
+            df_predictions["scale_trim"] = preprocessor.trim(
+                df_predictions["scale"], trim_min, trim_max, trim_tolerance
+            )
+            df_predictions["scale_trim_round"] = np.rint(df_predictions['scale_trim']).astype("int64")
+
+    # return the predictions as a dictionary but without the ID column
+    return df_predictions.drop("spkitemid", axis="columns").to_dict(orient="records")[0]
```
blongwill, line 187: Without the ID column because there’s only one sample being processed at a time, right?

rsmtool/rsmpredict.py
```diff
@@ -25,6 +28,165 @@
...
+        )
+        df_predictions["raw_trim_round"] = np.rint(df_predictions['raw_trim']).astype("int64")
+
+    # compute scaled predictions if the scaling parameters are available
+    if train_predictions_mean and train_predictions_sd and h1_mean and h1_sd:
+        df_predictions["scale"] = (
+            (df_predictions["raw"] - train_predictions_mean) / train_predictions_sd
+        ) * h1_sd + h1_mean
+
+        # apply trimming to scaled scores too, if possible
```
blongwill, line 179: Can we combine both of the trim max and min if statements on L167 and L180 into one? Once outside the scope of the if statement on L174, we can just iterate over the columns of df_predictions?

I was thinking something like this:

```
# now compute the raw prediction for the given features
    df_predictions = modeler.predict(df_processed_features)

    # compute scaled predictions if the scaling parameters are available
    if train_predictions_mean and train_predictions_sd and h1_mean and h1_sd:
         df_predictions["scale"] = (
             (df_predictions["raw"] - train_predictions_mean) / train_predictions_sd
         ) * h1_sd + h1_mean

    # trim predictions if ``trim_min`` and ``trim_max`` were specified
    if trim_min and trim_max:
        for score_proc in df_predictions.columns:

            df_predictions[f"{score_proc}_trim"] = preprocessor.trim(
                df_predictions[score_proc], trim_min, trim_max, trim_tolerance
            )
            df_predictions[f"{score_proc}_trim_round"] = np.rint(df_predictions[f'{score_proc}_trim']).astype("int64")
```

rsmtool/rsmpredict.py
desilinguist: Good catch. I'll fix this one but you are right in that they are used interchangeably throughout the codebase.

rsmtool/rsmpredict.py
desilinguist: It's because the ID was just a dummy one we added – it's not actually part of the input.

rsmtool/rsmpredict.py
desilinguist: Neat! However, we can't just use all of `df_predictions.columns` directly since that also has `spkitemid` at this point. We'll need to filter that out but that's still more readable, thanks!

desilinguist:
@blongwill  thanks for the great review! I have incorporated your suggestions. Please re-review and approve if everything looks good.

rsmtool/rsmpredict.py
blongwill: Is that dummy ID required to use preprocessor.trim? I'm wondering if it would be possible to make that drop earlier in the code at like L171 so that we don't have to add the second drop in the loop at L174

desilinguist: Ah, good call, `trim()` just needs the values and not the data frame itself. Fixed.

blongwill (approved):
Thanks for this! I added one last comment but you can definitely take it or leave it! Either way this is great!

mulhod (approved):

tazin-afrin (approved):
The new changes looks good to me, thanks!

amandameganchan (approved):

PR #589, opened 2023-01-14
Restore and increase test coverage

desilinguist:
After banging my head against this for a whole day, I finally figured out that adding a `pyproject.toml` disables the [measurement of test coverage in subprocesses](https://coverage.readthedocs.io/en/7.0.5/subprocess.html) for some inexplicable reason. This means that we lose coverage of the code in all the entry point scripts under the `main()` function. I tried lots of solutions but in the end I had to end up deleting `pyproject.toml` and just rely on other configuration files. 

Here are the actual changes in this PR:
- Delete `pyproject.toml`. 
- Add `.isort.cfg` to control isort during development.
- Update `.pre-commit-config.yaml` to include all of the flags that we need all the tools to have.
- Update the CI configuration files to no longer use the `defaults` channel or the `nose-cov` PyPI package since they are no longer needed. 
- Replace `coveralls` with `codecov` in `requirements.txt`.
- Add new tests for `convert_feature_json.py`.


mulhod:
Well, it's unfortunate that it's interacting in this way.

mulhod (approved):

desilinguist:
> Well, it's unfortunate that it's interacting in this way.

Yeah it's very strange. I tried lots of things like moving the `coverage` configuration from `.coveragerc` into `pyproject.toml`, keeping both around, etc. and nothing worked. Very strange.

Frost45 (approved):

PR #594, opened 2023-04-12
Fix bug in wording of what will be highlighted for disattenuated correlation

mulhod:


desilinguist (approved):
Thanks for spotting this!

tamarl08 (approved):

PR #595, opened 2023-05-15
Integrate SHAP into rsmtool for non-linear model explanations

damien2012eng:
This PR adds a new module called rsmexplain meant to integrate SHAP (Shapley Additive exPlanations) a library designed to offer explanations for model decisions through Shapley Values to the RSMTool. Users can provide a trained model, some background data, and some data for explanation. rsmexplain generates a report that includes tables, plot visualizations, and general information on their shap values. 

In addition to adding this new module, this PR also adds tests for it. Documentation and more plots will be updated in the next couple PRs.


damien2012eng:
Please consider this PR as a work-in-process PR. There are still a few parts under the development: 

1. To re-produce the test results for rsmexplain. I fixed the [random seed](https://github.com/EducationalTestingService/rsmtool/blob/shap_rsmt_develop/rsmtool/rsmexplain.py#L290) when creating the rsmexplainer. Is it the best practice to do it? 
2. The [auto-generate](https://github.com/EducationalTestingService/rsmtool/blob/shap_rsmt_develop/rsmtool/rsmexplain.py#L504-L507) is still under development. 
3. In the test_experiment_rsmexplain, I haven't include any tests to cover raising errors. I can do that as part of the current PR.

remo-help:
> * To re-produce the test results for rsmexplain. I fixed the [random seed](https://github.com/EducationalTestingService/rsmtool/blob/shap_rsmt_develop/rsmtool/rsmexplain.py#L290) when creating the rsmexplainer. Is it the best practice to do it?
You're using a Sampling explainer, which is based on this paper (https://dl.acm.org/doi/pdf/10.5555/1756006.1756007).
Since this method uses random feature permutations, the output should not be deterministic. Fixing the random seed, should be right approach here, I think. As long as the model you're using the fixed-seed explainer is quasi-deterministic in its outputs (like a linear regression for example), you should always get the same results.

If you're using a model that might give you a slight variation in outputs, then you need to make sure that the outputs on that side are reproducible. For example, if you use a[ Linear SVM](https://scikit-learn.org/stable/modules/generated/sklearn.svm.LinearSVR.html#sklearn.svm.LinearSVR), you need to fix the `random_stateint` param for reproducibility.

I see you're fixing the random seed in your unit test on the classifier side, so that should be the right way.

tests/test_experiment_rsmexplain.py
remo-help, line 1: you want to make sure here that all of these models you're using have a fixed random seed or their outputs are otherwise reproducible, if you want to test the SHAP values of these models at least

remo-help (approved):
I only took a brief look at the tests, with your question about reproducibility in mind.
Looks like the core of the module is pretty much still what I wrote, so I'll rubberstamp my past self here.
You can re-request a review if you change anything in `rsmexplain.py` that you want input on

rsmtool/utils/conversion.py
```diff
@@ -43,3 +45,40 @@ def convert_to_float(value):
...
+        start_index = string_to_number(groups.get("start"))
+        end_index = string_to_number(groups.get("end"))
+
+        # the range is valid if and only if:
+        # - both start and end indices are positive integers
+        # - and start index < end index
+        valid_range = (
+            isinstance(start_index, int)
+            and isinstance(end_index, int)
+            and 0 <= start_index < end_index
```
damien2012eng, line 78: Should we add an upper bound?

remo-help: if this does what I think it does, it probably makes more sense to evaluate whether the range exceeds the length of your feature set at the subsampling step. If your upper bound exceeds the total length of the set, just re-evaluate it to len(feature_set) and send a log message indicating that the range was larger than the set.
Otherwise you need to set an arbitrary upper limit or check the feature set length at this step here.

rsmtool/preprocessor.py
```diff
@@ -2614,6 +2609,107 @@ def process_data_rsmpredict(self, config_obj, data_container_obj):
...
+            If data contains duplicate response IDs.
+        """
+        df_background_features = data_container_obj.background_features
+        df_explainable_features = data_container_obj.explainable_features
+        df_feature_info = data_container_obj.feature_info
+
+        # get the column name that holds the IDs
+        id_column = config_obj["id_column"]
+
+        # should features be standardized?
```
damien2012eng, line 2648: Change this?

rsmtool/rsmexplain.py
```diff
@@ -199,25 +204,67 @@ def generate_explanation(
...
-        configuration["background_data"], configuration.configdir
+    experiment_ids = [splitext(basename(mf))[0] for mf in model_files]
+    if experiment_id not in experiment_ids:
+        raise FileNotFoundError(
+            f"{experiment_output_dir} does not contain a model "
+            f'for the experiment "{experiment_id}". The following '
+            f"experiments are contained in this directory: {experiment_ids}"
+        )
+
+    # check that the directory contains outher required files
```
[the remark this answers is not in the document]
desilinguist: You can fix this in your commit, @damien2012eng.

rsmtool/utils/conversion.py
desilinguist: We don't need an upper bound because even the endpoint is larger, the Python indexing will work just fine without raising an error. For example.

```
In [1]: a = list(range(1, 100))

In [2]: a[90:110]
Out[2]: [91, 92, 93, 94, 95, 96, 97, 98, 99]
```

rsmtool/preprocessor.py
```diff
@@ -2614,6 +2609,107 @@ def process_data_rsmpredict(self, config_obj, data_container_obj):
...
+        -------
+        config_obj : configuration_parser.Configuration
+            A new configuration object.
+        data_congtainer : container.DataContainer
+            A new data container object.
+
+        Raises
+        ------
+        KeyError
+            If columns specified in the configuration do not exist in the data.
```
damien2012eng, line 2637: Remove this KeyError?

rsmtool/preprocessor.py
desilinguist: This comment just indicates what `standardize_features` variable represents. So what's to change?

rsmtool/preprocessor.py
desilinguist: Yes, good call. You can remove this.

rsmtool/utils/conversion.py
desilinguist: We should try and add a test for this to be sure.

damien2012eng: Sure. I will add it.

desilinguist:
The latest changes look great @damien2012eng! @tamarl08 this is now ready for your review.

examples/rsmexplain/config_rsmexplain.json
```diff
@@ -0,0 +1,11 @@
+{
+    "description": "Explaning an linear regression model trained on all features.",
```
tamarl08, line 2:
```suggestion
    "description": "Explaning a linear regression model trained on all features.",
```

rsmtool/rsmexplain.py
```diff
@@ -0,0 +1,525 @@
...
+
+def select_features(featureset, range_size=None):
+    """
+    Sample examples from the given featureset and return indices.
+
+    Parameters
+    ----------
+    featureset: skll.data.FeatureSet
+        The SKLL FeatureSet object from which we are sampling.
+    range_size: Optional[int, Tuple[int, int]]
```
tamarl08, line 52:
```suggestion
    range_size: Optional[Union[int, Tuple[int, int]]]
```

rsmtool/rsmexplain.py
```diff
@@ -0,0 +1,525 @@
...
+from .preprocessor import FeaturePreprocessor
+from .reader import DataReader
+from .reporter import Reporter
+from .utils.commandline import ConfigurationGenerator, setup_rsmcmd_parser
+from .utils.constants import VALID_PARSER_SUBCOMMANDS
+from .utils.conversion import parse_range
+from .utils.logging import LogFormatter
+
+
+def select_features(featureset, range_size=None):
```
tamarl08, line 44: The name is confusing since no feature selection is done. Rename to something like `get_subset`?

rsmtool/rsmexplain.py
```diff
@@ -0,0 +1,525 @@
...
+    selected_feature_map = select_features(featureset, range_size=feature_range)
+
+    # if the user specified a sample size or a range, use it; otherwise all
+    # features will be selected
+    if feature_range:
+        positions = list(selected_feature_map.keys())
+        features = features[positions, :]
+
+    # convert to a dense array if not already one
+    features = features.toarray() if not isinstance(features, np.ndarray) else features
```
tamarl08, line 126: is the else required?

rsmtool/rsmexplain.py
```diff
@@ -0,0 +1,525 @@
...
+        A 2D numpy array containing sampled feature rows.
+    """
+    # get a sparse matrix with the features that were actually used
+    features = learner.feat_selector.transform(
+        learner.feat_vectorizer.transform(
+            featureset.vectorizer.inverse_transform(featureset.features)
+        )
+    )
+
+    # sample exampeles from the featureset
```
tamarl08, line 116:
```suggestion
    # sample examples from the featureset
```

rsmtool/rsmexplain.py
```diff
@@ -0,0 +1,525 @@
...
+        Defaults to ``None``.
+
+    Raises
+    ------
+    AssertionError
+        If the size of the background sample is < 300, to avoid inaccurate explanations.
+    FileNotFoundError
+        If any file contained in ``config_file_or_obj_or_dict`` cannot be located.
+    ValueError
+        If range is not an integer of list of integer.
```
tamarl08, line 165: what is `range`? is in the configuration?

tamarl08: looks like it raises an error only if both `sample_range` and `sample_size` are specified. not if none of them.

rsmtool/rsmexplain.py
```diff
@@ -0,0 +1,525 @@
...
+
+    experiment_ids = [splitext(basename(mf))[0] for mf in model_files]
+    if experiment_id not in experiment_ids:
+        raise FileNotFoundError(
+            f"{experiment_output_dir} does not contain a model "
+            f'for the experiment "{experiment_id}". The following '
+            f"experiments are contained in this directory: {experiment_ids}"
+        )
+
+    # check that the directory contains outher required files
```
tamarl08, line 243:
```suggestion
    # check that the directory contains other required files
```

rsmtool/rsmexplain.py
```diff
@@ -0,0 +1,525 @@
...
+        [configuration["background_data"], configuration["explainable_data"]],
+        configuration.configdir,
+    )
+    if not background_data_path:
+        raise FileNotFoundError(f"Input file {configuration['background_data']} does not exist")
+    if not explainable_data_path:
+        raise FileNotFoundError(f"Input file {configuration['explainable_data']} does not exist")
+
+    # read the background data, explainable data, and feature info files
+    feature_info = join(experiment_output_dir, f"{experiment_id}_feature.csv")
```
tamarl08, line 263:
```suggestion
    feature_info_path = join(experiment_output_dir, f"{experiment_id}_feature.csv")
```

rsmtool/rsmexplain.py
```diff
@@ -0,0 +1,525 @@
...
+        "feature_info",
+    ]
+    reader = DataReader(file_paths, file_names)
+    container = reader.read(kwargs_dict={"feature_info": {"index_col": 0}})
+
+    # ensure that the background data is large enough for meaningful explanations
+    background_data_size = len(container.background_features)
+    try:
+        assert background_data_size > 300
+    except AssertionError:
```
tamarl08, line 277: 1. Wouldn't it be better to use a simple if and raise an error?
2. currently it just exits, without raising the error (as stated in the feature docstring).
3. should it be >= 300?

rsmtool/rsmexplain.py
desilinguist: yeah that's fair. May be `select_examples` instead of `select_features` ?

rsmtool/rsmexplain.py
desilinguist: All good points. Will fix!

rsmtool/rsmexplain.py
```diff
@@ -0,0 +1,525 @@
...
+
+    Parameters
+    ----------
+    explanation: shap.Explanation
+        SHAP explanation object containing shap values, data points, feature
+        names and base values.
+    output_dir : str
+        Path to the experiment output directory.
+    ids: dict
+        Dictionary mapping new row indices to original FeatureSet indices.
```
tamarl08, line 383:
```suggestion
        Dictionary mapping new row indices to original FeatureSet ids.
```

rsmtool/rsmexplain.py
```diff
@@ -0,0 +1,525 @@
...
+    logger = logger if logger else logging.getLogger(__name__)
+
+    # get the various output sub-directories which should already exist
+    csvdir = abspath(join(output_dir, "output"))
+    reportdir = abspath(join(output_dir, "report"))
+
+    # get the experiment ID
+    experiment_id = configuration["experiment_id"]
+
+    # first write the explanation object to disk, in case need it later
```
tamarl08, line 400:
```suggestion
    # first write the explanation object to disk, in case we need it later
```

rsmtool/rsmexplain.py
```diff
@@ -0,0 +1,525 @@
...
+    ).transpose()
+    df_abs.to_csv(csv_path_abs, index_label="Feature")
+
+    # Initialize a reporter instance and add the sections:
+    reporter = Reporter(logger=logger)
+    general_report_sections = configuration["general_sections"]
+    special_report_sections = configuration["special_sections"]
+
+    # get any custom sections and locate them to make sure
+    # that they exist, otherwise raise an exception
```
tamarl08, line 433: no exception is raised - will in come from the Reporter call?

rsmtool/preprocessor.py
```diff
@@ -2614,6 +2609,105 @@ def process_data_rsmpredict(self, config_obj, data_container_obj):
...
+            None,
+            None,
+            None,
+            None,
+            None,
+        )
+
+        # check that the id_column in both frames contains unique values
+        if (
+            df_background_preprocessed["spkitemid"].size
```
tamarl08, line 2674: what is `spkitemid` comparing to `id_column`?

rsmtool/rsmexplain.py
desilinguist: yep, that's right.

rsmtool/preprocessor.py
```diff
@@ -2614,6 +2609,105 @@ def process_data_rsmpredict(self, config_obj, data_container_obj):
...
+        df_feature_info = data_container_obj.feature_info
+
+        # get the column name that holds the IDs
+        id_column = config_obj["id_column"]
+
+        # should features be standardized?
+        standardize_features = config_obj.get("standardize_features", True)
+
+        # rename the ID columns in both frames
+        df_background_preprocessed = self.rename_default_columns(
```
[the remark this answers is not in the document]
desilinguist: Lot of the analyses assume that ID and score columns are named a specific thing. That's how we implemented it to make things easier.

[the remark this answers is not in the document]
tamarl08: so the id provided in the config is not necessarily one of the columns in the data?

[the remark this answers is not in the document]
desilinguist: Sorry, I updated the comment.

rsmtool/preprocessor.py
tamarl08: the check is done on `spkitemid` while the error message saying `column_id`. 
so `column id` was renamed to `spkitemid` in the previous step?

desilinguist: Yes, at this point, `id_column` has been renamed to `spkitemid`.

desilinguist: I will fix the commend though.

tamarl08: I don't think you need to fix it, if it's the column name the user provided.

tamarl08: For readability, it  could be a global variable e.g. DEFAULT_ID="spkitemid"

rsmtool/notebooks/explanations/auto_cohorts.ipynb
```diff
@@ -0,0 +1,106 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "5a4205dc",
+   "metadata": {},
+   "source": [
+    "#### Auto cohort bar plots\n",
+    "\n",
+    "This plot represents a bar plot split into N cohorts that optimally separate the SHAP values of the exammples using a `DecisionTreeRegressor` from scikit-learn. The plots show the cohorts using differently shaded bars and the legend indicates the metric along which the cohorts were chosen. The features are plotted using absolute mean impact.\n",
```
tamarl08, line 10:
```suggestion
    "This plot represents a bar plot split into N cohorts that optimally separate the SHAP values of the examples using a `DecisionTreeRegressor` from scikit-learn. The plots show the cohorts using differently shaded bars and the legend indicates the metric along which the cohorts were chosen. The features are plotted using absolute mean impact.\n",
```

rsmtool/notebooks/explanations/shap_values.ipynb
```diff
@@ -0,0 +1,207 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "### Reading Shap values\n",
+    "\n",
+    "Shap values are additive representations of a feature's impact on a model decision. The sum of all shap values and the base value for a example yields the actual model prediction for that example.\n",
```
tamarl08, line 31:
```suggestion
    "Shap values are additive representations of a feature's impact on a model decision. The sum of all shap values and the base value for an example yields the actual model prediction for that example.\n",
```

rsmtool/notebooks/explanations/shap_values.ipynb
```diff
@@ -0,0 +1,207 @@
...
+    "\n",
+    "A shap value for a feature can be considered that feature's contribution to that specific prediction. By computing the averge of all absolute shap values for a specific feature, we can calculate that feature's average impact on the predictions for the data we are trying to explain. The average, mximum, and minimum absolute shap values for all features can be found in the `<id>_absolute_shap_values.csv` file in the `output` directory.\n",
```
tamarl08, line 33:
```suggestion
    "A shap value for a feature can be considered that feature's contribution to that specific prediction. By computing the averge of all absolute shap values for a specific feature, we can calculate that feature's average impact on the predictions for the data we are trying to explain. The average, maximum, and minimum absolute shap values for all features can be found in the `<id>_absolute_shap_values.csv` file in the `output` directory.\n",
```

rsmtool/notebooks/explanations/shap_values.ipynb
```diff
@@ -0,0 +1,207 @@
...
+   "source": [
+    "mean_values = df_abs_shap[\"abs. mean shap\"]\n",
+    "rows_with_zero_mean_values = df_abs_shap[df_abs_shap[\"abs. mean shap\"] == 0]\n",
+    "\n",
+    "msg = (\"#### Features with zero mean Shap value\\n The featuers in table below \"\n",
+    "       \"likely did not contribute to the model decisions. Assuming the examples passed \"\n",
+    "       \"were sufficient in number and representative of the data the model usually \"\n",
+    "       \"encounters, the features in this table are not useful to the model.\\n \"\n",
+    "       \"**IMPORTANT**: Before you draw this conclusion, please make sure that those \"\n",
+    "       \"features were not simply 0 in all the examples you provided the model.\")\n",
```
tamarl08, line 111: `in all the examples you provided the model` - does that mean in the training data?

tamarl08: Also, maybe this section should come after the `Bottom 5 features by mean absolute Shap value`?

tamarl08 (changes requested):
@damien2012eng @desilinguist I finished reviewing the files, looks good! had only few comments. After reviewing the documentation I will also try to use it and generate the reports.

rsmtool/notebooks/explanations/shap_values.ipynb
desilinguist: Yes, I think it means training data. Although I suppose it was all zero in the background or explainable data, it would also not come up as very important?

desilinguist:
Thank you so much @tamarl08! I will address all your comments and let you know. Once you approve we can merge this, rebase the documentation PR, and then you can start reviewing that one.

rsmtool/notebooks/explanations/shap_values.ipynb
tamarl08: I think it may not be the training data.

> the value for a feature indicates, for every model decision passed to the explainer, how the model decision would be impacted by removing that feature.

possibly shap removes the feature by setting it to 0?

desilinguist: I don't think it would do it that way because then how would it distinguish the case where a feature was actually zero in all the instances?

tamarl08: it would add a comment like the above...

remo-help: Features that have a 0 mean shap value did not contribute to the model decisions in all rows that were passed. If you passed a representative sample, then that means that this feature is not useful to your model. 

However, if the sample is not representative, this could be misleading:
So imagine you train a model on 10k rows. You pass 20 rows to a SHAP explanation. Coincidentally, in those 20 rows a feature that usually has an impact does not have an impact for those 20 decisions, for whichever reason. That's why that disclaimer is there.

So this is not about training data for the model, but about the data you pass to SHAP.

hope that helps

desilinguist: Thanks @remo-help  I think that makes sense. So, the disclaimer is about whether the sample you are passing to SHAP to explain is representative of your real training data or not. If the real training data also had that feature as always zero, your representative explain data will have it too. It's only about when the two distributions don't match up. I will clarify.

remo-help: The real training data might have actual values for that feature, but your model may have decided to ignore that feature altogether. That would also get you a mean SHAP value of 0. This is pretty unlikely of course, but in theory possible.

desilinguist: Yeah that would be pretty unlikely. In any case, I will make the disclaimer clearer.

desilinguist:
@tamarl08 the PR is ready for re-review.

rsmtool/notebooks/explanations/shap_values.ipynb
```diff
@@ -152,6 +153,39 @@
...
+   "execution_count": null,
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "mean_values = df_abs_shap[\"abs. mean shap\"]\n",
+    "rows_with_zero_mean_values = df_abs_shap[df_abs_shap[\"abs. mean shap\"] == 0]\n",
+    "\n",
+    "msg = (\"#### Features with zero mean Shap value\\n The features in table below \"\n",
```
tamarl08, line 167: The features in table below ->  The features in **the** table below
(for some reason I don't have the option to edit in the text)

tamarl08 (approved):

PR #601, opened 2023-06-09
Add rsmexplain documentation and update RTD requirements.

desilinguist:
- Add `shap` to `doc/requirements.txt`.
- Add documentation for `rsmexplain`.


desilinguist:
@tamarl08  this is ready for review.

doc/config_rsmexplain.rst.inc
```diff
@@ -0,0 +1,66 @@
...
+
+.. note:: To make it easy to get started with  ``rsmexplain``, we provide a way to **automatically generate** configurations file both interactively as well as non-interactively. Novice users will find interactive generation more helpful while more advanced users will prefer non-interactive generation. See :ref:`this page <autogen_configuration>` for more details.
+
+Next, we describe all of the ``rsmexplain`` configuration fields in detail. There are four required fields and the rest are all optional. We first describe the required fields and then the optional ones (sorted alphabetically).
+
+background_data
+~~~~~~~~~~~~~~~
+The path to the background data feature file one of the :ref:`supported formats <input_file_format>`. Each row should correspond to a single response and contain the numeric feature values extracted for this response. In addition, there should be a column containing a unique identifier (ID) for each response. This path can be absolute or relative to the location of config file. It must contain at least 300 responses to ensure meaningful explanations.
```
tamarl08, line 14:
```suggestion
The path to the background data feature file in one of the :ref:`supported formats <input_file_format>`. Each row should correspond to a single response and contain the numeric feature values extracted for this response. In addition, there should be a column containing a unique identifier (ID) for each response. This path can be absolute or relative to the location of the config file. It must contain at least 300 responses to ensure meaningful explanations.
```

doc/config_rsmexplain.rst.inc
```diff
@@ -0,0 +1,66 @@
...
+~~~~~~~~~~~~~
+An identifier for the ``rsmexplain`` experiment. This will be used to name the report. It can be any combination of alphanumeric values, must *not* contain spaces, and must *not* be any longer than 200 characters.
+
+experiment_dir
+~~~~~~~~~~~~~~
+The directory containing the rsmtool models we want to explain. This directory should contain an ``output`` sub-directory and that sub-directory should contain two files: the ``<experiment_id>.model`` and  ``<experiment_id>_feature.csv``. Note that ``<experiment_id>`` refers to the one defined in this same configuration file. As an example of this directory structure, take a look at the ``existing_experiment`` directory  `here <https://github.com/EducationalTestingService/rsmtool/tree/main/tests/data/experiments/svr-explain>`__
+
+background_kmeans_size *(Optional)*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The size of the k-means sample for background. Defaults to ``500``. A dataset is summarized with k mean samples weighted by the number of data points they each represent.
```
tamarl08, line 30: Is there a rule-of-thumb for setting this number based on data characteristics?

doc/config_rsmexplain.rst.inc
```diff
@@ -0,0 +1,66 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~
+If we want to explain a random sample of the responses in explainable_data, we can specify the size of that random sample here. For example, to explain a random sample of 10 responses, we would set this to 10. Defaults to None.
+
+.. note ::
+
+    Only one of ``sample_range`` or ``sample_size`` must be specified. If neither is specified, explanations will be generated for the entire set of responses in ``explainable_data`` which could be very slow, depending on its size.
+
+show_auto_cohorts *(Optional)*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If this option is set to ``true``, auto cohort bar plots will be displayed. These plots can be useful to detect interaction effects between cohorts and features. If a cohort shows a high feature value, then there may be an interaction between that cohort and the feature. Defaults to ``false``.
```
tamarl08, line 62: why is this optional? is it very slow?

doc/config_rsmexplain.rst.inc
```diff
@@ -0,0 +1,66 @@
...
+
+standardize_features *(Optional)*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If this option is set to ``false``, the feature values for the responses in ``background_data`` and ``explainable_data`` will *not* be standardized using the mean and standard deviation parameters from the rsmtool experiment from ``<experiment_dir>/output/<experiment_id>_feature.csv``. This should only be set to ``false`` if the same option was also set to ``false`` when running the original rsmtool experiment. Defaults to ``true``.
```
tamarl08, line 66: if it was set to false in the original experiment, it *must* be set to false here?
can we tell that this was the case from the original experiment output?

doc/config_rsmexplain.rst.inc
```diff
@@ -0,0 +1,66 @@
...
+
+.. note:: To make it easy to get started with  ``rsmexplain``, we provide a way to **automatically generate** configurations file both interactively as well as non-interactively. Novice users will find interactive generation more helpful while more advanced users will prefer non-interactive generation. See :ref:`this page <autogen_configuration>` for more details.
+
+Next, we describe all of the ``rsmexplain`` configuration fields in detail. There are four required fields and the rest are all optional. We first describe the required fields and then the optional ones (sorted alphabetically).
+
+background_data
+~~~~~~~~~~~~~~~
+The path to the background data feature file one of the :ref:`supported formats <input_file_format>`. Each row should correspond to a single response and contain the numeric feature values extracted for this response. In addition, there should be a column containing a unique identifier (ID) for each response. This path can be absolute or relative to the location of config file. It must contain at least 300 responses to ensure meaningful explanations.
+
+explainable_data
```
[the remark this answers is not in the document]
tamarl08: I see what you mean. `explainable_data` sounds to me like there is some requirement on the data, that it will be "explainable". but if this is the naming used by shap, that's fine.

doc/usage_rsmexplain.rst.inc
```diff
@@ -0,0 +1,59 @@
+.. _usage_rsmexplain:
+
+``rsmexplain`` - Explain non-linear models
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+RSMTool provides the ``rsmexplain`` command-line utility to generate a report to explain the predictions made by a model trained using the ``rsmtool`` utility. These explanations contain useful information about the contribution of each feature to the final score, even if the model is non-linear or black-box in nature.
```
tamarl08, line 6:
```suggestion
RSMTool provides the ``rsmexplain`` command-line utility to generate a report explaining the predictions made by a model trained using the ``rsmtool`` utility. These explanations contain useful information about the contribution of each feature to the final score, even if the model is non-linear or black-box in nature.
```

doc/usage_rsmexplain.rst.inc
```diff
@@ -0,0 +1,59 @@
...
+
```
tamarl08, line 7: Add a reference to shap documentation in this introduction.

doc/usage_rsmexplain.rst.inc
```diff
@@ -0,0 +1,59 @@
...
+``rsmexplain`` uses the sampling explainer which is model agnostic and should, in principle, work for any type of model. However, ``rsmexplain`` currently only supports regressors since they are the most popular model type used for automated scoring.
```
tamarl08, line 8: This comment could be in a Note so users won't miss it.

doc/config_rsmexplain.rst.inc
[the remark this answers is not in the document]
desilinguist: I don't think SHAP has the convention to call this `explainable`. I think this is a name we just chose. I agree with @tamarl08  that `explainable` does imply something other than "this is the data we want to explain". How about `explained_data`?

doc/config_rsmexplain.rst.inc
desilinguist: @remo-help I think you did a bunch of experiments to come up with this number or is this recommended by SHAP? Adding some details as to how we came up with this number would be useful, I agree.

doc/config_rsmexplain.rst.inc
desilinguist: Good question, actually. May be it's because they aren't very helpful if you have a lot of features? @remo-help can you shed some light on why we made this optional? I can't remember now.

doc/config_rsmexplain.rst.inc
desilinguist: We can if we have the full output of the original experiment including the saved configuration file. I guess we should try to see if we have that available and try to read it, if it is.

doc/config_rsmexplain.rst.inc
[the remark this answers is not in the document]
tamarl08: yes `explained_data` is better.  how about `explain_data`?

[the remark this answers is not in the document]
desilinguist: It's definitely shorter but it kind of breaks the cognitive alignment with `background_data by using an imperative instead of an adjective as the first word? May be I am just overthinking it?

doc/config_rsmexplain.rst.inc
tamarl08: I think a requirement to run this following an rsmtool experiemnt with all its input and output data in the expected format is reasonable, we already assume part of it.

doc/config_rsmexplain.rst.inc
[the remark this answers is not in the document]
tamarl08: Overthinking here too, explained_data suggests that it has already been explained.
`explanation_data`?

(but really, all the previous suggestions are fine)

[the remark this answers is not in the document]
desilinguist: Yeah, I agree. `explain_data` sounds the best out of all the options to me. @damien2012eng can you please rename `explainable_data` to `explain_data` in all the code and documentation?

doc/config_rsmexplain.rst.inc
desilinguist: Fair point. I created an issue #602 . @damien2012eng  do you feel comfortable taking a stab at this?

doc/config_rsmexplain.rst.inc
remo-help: The reason I set this to optional initially was because this feature can be unstable or provide misleading information if the users do not pass a large dataset for explanation 
I think it was also sometimes a pretty wonky if it could not establish clear cohorts, because the way this works is that it tries to automatically detect possible cohorts (think something like: "rows where feature 24 is > .56 ") and if there are no strong interactions the graphs just look weird

remo-help: it should be unrelated to the total amount of features, since you can still potentially get strong interaction effects between features even if you have a lot, but it might become less likely

doc/config_rsmexplain.rst.inc
desilinguist: This is very helpful! Thanks @remo-help. @damien2012eng let's add this information.

tamarl08: Thanks @remo-help! 
Is there a threshold for the sample size above which the running time is significantly slow?
Also can we recommend not to use sampling when the background data size is lower than some threshold (500 like the current default or maybe more)?

tamarl08: Thanks for the detailed explanation @remo-help!

@damien2012eng @desilinguist we say here that it's an optional value but we use it with no default here: https://github.com/EducationalTestingService/rsmtool/blob/main/rsmtool/rsmexplain.py#L307

Also we cluster on the masked dataset, meaning if the user specified sample size or range,  the input for kmeans will be on that sampled data.

desilinguist: Good catch on the optional. 

I think that's how @remo-help had implemented it, IIRC. @remo-help please let us know if masking the featureset _before_ doing k-means clustering is the wrong thing to do?

desilinguist: Actually, we are calling `mask()` without the `feature_range` parameter so that means all of the features are selected. We still call `mask()` since it also filters out feature columns that didn't end up being used by the model.

tamarl08: right! yes that seems correct.

remo-help: @desilinguist  that's correct, `mask()` also makes sure the array is of the right type.
@tamarl08 in the line you reference, I'm pretty sure the default is specified in the config file, so if the user doesnt specify anything, the value pulled here will automatically be the default

desilinguist: yep, @remo-help  is right. The default value is automatically set when the configuration is parsed and nothing is specified by the user.

tamarl08: can we set the default to min(background size, 500)?
I can't find the correct `shap.kmeans` method, it may be ok to pass the a `k` larger than `n`.

doc/tutorial_rsmexplain.rst.inc
```diff
@@ -0,0 +1,80 @@
...
+.. literalinclude:: ../examples/rsmexplain/config_rsmexplain.json
+    :language: javascript
+    :linenos:
+
+Let's take a look at the options in our configuration file.
+
+- **Line 2**: We give a short description of this experiment. This will be shown in the report.
+- **Line 3**: We give the path to the directory containing the output of the original `rsmtool`` experiment. Note that this is the top-level directory that contains the ``output`` directory produced by ``rsmtool``.
+- **Line 4**: We provide the experiment_id of the rsmtool experiment used to train the model. This can usually be read off the output/<experiment_id>.model file in the rsmtool experiment output directory.
+- **Line 5**: We list the path to the data file for use as the background distribution.
```
tamarl08, line 45:
```suggestion
- **Line 5**: We provide the path to the data file to be used as the background distribution.
```

doc/config_rsmexplain.rst.inc
desilinguist: Here's the method: https://github.com/slundberg/shap/blob/master/shap/utils/_legacy.py#L10

It's calling `KMeans` from scikit-learn. I expect it would raise an error if this were the case. @damien2012eng  can we test this by setting the kmeans parameter to a value larger than the background data size?

remo-help: `shap.kmeans()` is inconveniently not present in the SHAP documentation, although they use it in the examples.
I wouldn't be surprised if it was a simple wrapper around the `sklearn.cluster.KMeans` function: https://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html#sklearn.cluster.KMeans

doc/tutorial_rsmexplain.rst.inc
```diff
@@ -0,0 +1,80 @@
...
+- **Line 6**: We list the path to the data file to explain.
+- **Line 7**: This field indicates that the unique IDs for the responses in both data files are located in a column named ``ID``.
+- **Line 8**: This field indicates that we wish to explain a random sample of responses from the second data file of the given size.
```
tamarl08, line 48:
```suggestion
- **Line 8**: This field indicates that we wish to explain a random sample of 10 responses from the second data file.
```

doc/tutorial_rsmexplain.rst.inc
```diff
@@ -0,0 +1,80 @@
...
+    :language: javascript
+    :linenos:
+
+Let's take a look at the options in our configuration file.
+
+- **Line 2**: We give a short description of this experiment. This will be shown in the report.
+- **Line 3**: We give the path to the directory containing the output of the original `rsmtool`` experiment. Note that this is the top-level directory that contains the ``output`` directory produced by ``rsmtool``.
+- **Line 4**: We provide the experiment_id of the rsmtool experiment used to train the model. This can usually be read off the output/<experiment_id>.model file in the rsmtool experiment output directory.
+- **Line 5**: We list the path to the data file for use as the background distribution.
+- **Line 6**: We list the path to the data file to explain.
```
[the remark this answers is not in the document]
desilinguist: How about "We provide the path to the data file that will be used to generate the explanations."

tamarl08:
One last comment following the tutorial (which was clear and easy to run!):
I think the `Bottom 5 features` table should be sorted from lower score to higher, so the least meaningful feature is first. also, the numbers in that table should be rounded. See snapshot: 

Screenshot 2023-06-12 at 2 05 43 PM

doc/config_rsmexplain.rst.inc
```diff
@@ -13,7 +13,7 @@ background_data
 ~~~~~~~~~~~~~~~
 The path to the background data feature file in one of the :ref:`supported formats <input_file_format>`. Each row should correspond to a single response and contain the numeric feature values extracted for this response. In addition, there should be a column containing a unique identifier (ID) for each response. This path can be absolute or relative to the location of the config file. It must contain at least 300 responses to ensure meaningful explanations.
 
-explainable_data
+explain_data
 ~~~~~~~~~~~~~~~~
```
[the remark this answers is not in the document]
damien2012eng: Good catch!

doc/config_rsmexplain.rst.inc
damien2012eng: @desilinguist  is right. Here is the raised error from the sklearn when I set the kmeans_size to be `750`. 

> ValueError: n_samples=500 should be >= n_clusters=750.

damien2012eng:
@tamarl08 I fixed the sorting issue. It looks like the rounding is disabled on purpose. It is mentioned on the above of the chart: 

>  Rounding is disabled for this table to prevent values appearing as 0.

@remo-help could share more insights to us.

desilinguist:
> Rounding is disabled for this table to prevent values appearing as 0.

Yeah, I remember this now. I wonder if we can just display those numbers in scientific notation with 3 significant digits so they are not rounded but a bit easier to read?

desilinguist:
@tamarl08 does the scientific notation idea sound interesting?

desilinguist:
@tamarl08 actually no, I was just going to do that specific table in scientific notation no matter what and update the text to say why we are doing that.

desilinguist:
I can do the conditional formatting if you think that would be more useful.

desilinguist:
@tamarl08 I added the conditional formatting you were thinking of and @damien2012eng has also incorporated many other suggestions. Please re-review.

doc/config_rsmexplain.rst.inc
```diff
@@ -27,7 +27,9 @@ The directory containing the rsmtool models we want to explain. This directory s
 
 background_kmeans_size *(Optional)*
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-The size of the k-means sample for background. Defaults to ``500``. A dataset is summarized with k mean samples weighted by the number of data points they each represent.
+The size of the k-means sample for background sampling. Defaults to 500. We summarize the dataset specified in `background_data` with this many k-means clusters (each sluter is weighted by the number of data points it represents) and then use the summarized data set for sampling *instead* of the original. The k-means clustering allows us to speed up the explanation process but may sacrifice some accuracy. The default value of 500 has been shown to provide a good balance between speed and accuracy in our experiments. You may use a higher value if you have a very large or very diverse background dataset and you want to ensure that it's accurately summarized.
```
tamarl08, line 30:
```suggestion
The size of the k-means sample for background sampling. Defaults to 500. We summarize the dataset specified in `background_data` with this many k-means clusters (each cluster is weighted by the number of data points it represents) and then use the summarized data set for sampling *instead* of the original. The k-means clustering allows us to speed up the explanation process but may sacrifice some accuracy. The default value of 500 has been shown to provide a good balance between speed and accuracy in our experiments. You may use a higher value if you have a very large or very diverse background dataset and you want to ensure that it's accurately summarized.
```

doc/usage_rsmexplain.rst.inc
```diff
@@ -0,0 +1,61 @@
+.. _usage_rsmexplain:
+
+``rsmexplain`` - Explain non-linear models
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+RSMTool provides the ``rsmexplain`` command-line utility to generate a report explaining the predictions made by a model trained using the ``rsmtool`` utility. These explanations contain useful information about the contribution of each feature to the final score, even if the model is non-linear or black-box in nature. The `rsmxxplain` module utilizes the `SHAP <https://shap.readthedocs.io/en/latest/index.html>_` library for the explanability.
```
tamarl08, line 6:
```suggestion
RSMTool provides the ``rsmexplain`` command-line utility to generate a report explaining the predictions made by a model trained using the ``rsmtool`` utility. These explanations contain useful information about the contribution of each feature to the final score, even if the model is non-linear or black-box in nature. The `rsmexplain` module utilizes the `SHAP <https://shap.readthedocs.io/en/latest/index.html>_` library for the explanability.
```

desilinguist:
Thanks for testing this @tamarl08! I think zero-ing out features in the explain data but keeping them as is in the background data can have the effect of making the feature look more important than it actually is. What SHAP is doing to "omit" a feature is simply replacing it with a sampled value from the background data and then measuring the difference in prediction. So, if the feature is always zero in the explain data, then the difference between an instance's assigned (incorrect) prediction and the perturbed prediction probably appears larger on average. 

Not sure about the pandas warning. Let me look into it a bit.

remo-help:
@tamarl08  Nitin is right, this is likely an accident of setting the feature to 0. SHAP gives you a local explanation of a model decision, and you get a global estimate by running a lot of decisions.
But, if you selectively modify one feature (by say setting it to 0), you give it unnatural data and you might end up making that feature more important to the model explanation. As Nitin said, the explainer does not literally set a feature to 0, rather it looks at the background set and tries to find something like an average or representative value for this feature. So, if you set the feature to an actual 0 value, that might be a highly unusual value for this feature, which will end up making it more influential for a model decision

rsmtool/notebooks/explanations/data_description.ipynb
```diff
@@ -39,17 +39,21 @@
...
+    "\n",
+    "However, running `rsmexplain` on very large background datasets can be computationally expensive and time-consuming. One trick we can use is to use k-means clustering to \"summarize\" and reduce the size of the background dataset without losing too much information. To do this, we run a k-means clustering algorithm on the background dataset which yields `k` clusters of feature values, each represented by a \"centroid\" (the average of all data points in the cluster). These centroids are then used to compute the SHAP values, i.e., when we \"omit\" a feature and go to replace its value, we do so by sampling from from one of the centroids rather than the original dataset. \n",
```
damien2012eng, line 54:
```suggestion
    "However, running `rsmexplain` on very large background datasets can be computationally expensive and time-consuming. One trick we can use is to use k-means clustering to \"summarize\" and reduce the size of the background dataset without losing too much information. To do this, we run a k-means clustering algorithm on the background dataset which yields `k` clusters of feature values, each represented by a \"centroid\" (the average of all data points in the cluster). These centroids are then used to compute the SHAP values, i.e., when we \"omit\" a feature and go to replace its value, we do so by sampling from one of the centroids rather than the original dataset. \n",
```

damien2012eng (approved):
Looks great!

rsmtool/notebooks/explanations/data_description.ipynb
```diff
@@ -39,17 +39,21 @@
...
+    "\n",
+    "`rsmexplain` applies this trick by default. The number of cluters `k` can be specified via the \"background_kmeans_size\" option in its configuration file. The default value is 500 since that has been shown to represent a good compromise between accuracy and speed."
```
tamarl08, line 56:
```suggestion
    "`rsmexplain` applies this trick by default. The number of clusters `k` can be specified via the \"background_kmeans_size\" option in its configuration file. The default value is 500 since that has been shown to represent a good compromise between accuracy and speed."
```

rsmtool/notebooks/explanations/data_description.ipynb
```diff
@@ -39,17 +39,21 @@
...
+    "The background data is responsible for generating the *base value* of your SHAP explanations. Specifically, this is the average model prediction in the background data passed to `rsmexplain`.\n",
     "\n",
-    "By default, rsmexplain generates a kmeans-clustered representation of your background sample. The size of which you can specify in the \"background_kmeans_size\" parameter. Generally a kmeans sample of size 100 or higher is sufficient to generate an accurate base value.\n"
+    "SHAP computes feature contributions by replacing feature values in the data you wish to explain with values sampled from the background set and measuring changes in the prediction. Therefore, The background data is crucial in helping us understand the impact of a feature moving from its 'baseline' or 'average' state to the current state. This movement is what the SHAP values shown in the report below quantify. This means that as long as **sufficiently large and diverse** background data is used, the SHAP values for individual examples in the explain data can be considered reliable.\n",
```
tamarl08, line 52:
```suggestion
    "SHAP computes feature contributions by replacing feature values in the data you wish to explain with values sampled from the background set and measuring changes in the prediction. Therefore, the background data is crucial in helping us understand the impact of a feature moving from its 'baseline' or 'average' state to the current state. This movement is what the SHAP values shown in the report below quantify. This means that as long as **sufficiently large and diverse** background data is used, the SHAP values for individual examples in the explain data can be considered reliable.\n",
```

tamarl08 (approved):

rsmtool/utils/notebook.py
```diff
@@ -30,14 +30,21 @@ def float_format_func(num, prec=3):
     prec: int, optional
         The number of decimal places to use when displaying the number.
         Defaults to 3.
+    scientific: bool, optional
+        Whether to display the number in in scientific notiation if
```
tamarl08, line 34:
```suggestion
        Whether to display the number in scientific notation if
```

desilinguist:
@tamarl08, @damien2012eng I noticed that there's a bug in how the bottom 5 table is produced. Fixing it in my next commit.

rsmtool/utils/notebook.py
```diff
@@ -30,14 +30,21 @@ def float_format_func(num, prec=3):
     prec: int, optional
         The number of decimal places to use when displaying the number.
         Defaults to 3.
+    scientific: bool, optional
+        Whether to display the number in scientific notiation if
```
tamarl08, line 34:
```suggestion
        Whether to display the number in scientific notation if
```

tamarl08:
@desilinguist all looks good including the report!

rsmtool/utils/notebook.py
desilinguist: @damien2012eng  can you fix this in your PR? I don't feel like running another build unless you also find some more typos? :)

damien2012eng (approved):

PR #617, opened 2023-07-16
Add wandb logging to rsmtool

tamarl08:
Add option to log to W&B from rsmtool.py.

Note: I didn't add this to unit tests since it only allows a very shallow testing. Instead I added a tutorial config that logs to wandb and verified that everything works.
See the results of the most recent run [here](https://wandb.ai/etslabs/rsmtool-tutorial/runs/lnv90fh7) or try it yourself!


doc/config_rsmtool.rst.inc
```diff
@@ -369,3 +369,23 @@ If set to ``true``, use the ``min`` and ``max`` columns specified in the ``featu
 .. note::
 
     If ``use_truncation_thresholds`` is set, a ``features`` file _must_ be specified, and this file _must_ include ``min`` and ``max`` columns. If no ``feature`` file is specified or these columns are missing, an error will be raised.
+
+.. _use_wandb:
+
+use_wandb *(Optional)*
+""""""""""""""""""""""
+If set to ``true``, the generated reports and all intermediate tables will be logged to `Weights & Biases <https://wandb.ai/>`__.
+The Weights & Biases entity and project name should be specified in the appropriate configuration fields. Defaults to ``false``.
```
desilinguist, line 378: Link "entity" and "project name" to the fields below using `:ref:`?

doc/usage_rsmtool.rst
```diff
@@ -43,6 +43,8 @@ Output
 ------
 
 ``rsmtool`` produces a set of folders in the experiment output directory. This is either the current directory in which ``rsmtool`` is run or the directory specified as the second optional command-line argument.
+If logging to `Weights & Biases <https://wandb.ai/>`__ is enabled, the reports and all intermediate files are also logged to
```
desilinguist, line 46: Link "enabled" to `use_wandb`?

doc/config_rsmtool.rst.inc
```diff
@@ -369,3 +369,23 @@ If set to ``true``, use the ``min`` and ``max`` columns specified in the ``featu
...
+
+.. _wandb_project:
+
+wandb_project *(Optional)*
+""""""""""""""""""""""""""
+The Weights & Biases project name if logging to Weights and Biases is enabled.
```
mulhod, line 384: Let's call it either "Weights & Biases" consistently.
```suggestion
The Weights & Biases project name if logging to Weights & Biases is enabled.
```

doc/config_rsmtool.rst.inc
```diff
@@ -369,3 +369,23 @@ If set to ``true``, use the ``min`` and ``max`` columns specified in the ``featu
...
+
+.. _wandb_entity:
+
+wandb_entity *(Optional)*
+""""""""""""""""""""""""""
+The Weights & Biases entity name if logging to Weights and Biases is enabled.
```
mulhod, line 390:
```suggestion
The Weights & Biases entity name if logging to Weights & Biases is enabled.
```

rsmtool/configuration_parser.py
```diff
@@ -1058,7 +1050,18 @@ def validate_config(cls, config, context="rsmtool"):
...
-        # 15. Clean up config dict to keep only context-specific fields
+        # 15. Check that if the user enables logging to W&B, they also
+        # specified wandb project and entity.
+        if new_config["use_wandb"]:
+            # make sure we have project name and entity
+            if "wandb_project" not in new_config or "wandb_entity" not in new_config:
+                raise ValueError(
+                    "You must specify both `wandb_project` "
+                    "and `wandb_entity` if you want to enable "
+                    "logging to W&B"
```
desilinguist, line 1061:
```suggestion
                    "logging to W&B."
```

doc/tutorial.rst
```diff
@@ -80,6 +80,10 @@ Documentation for all of the available configuration options is available :ref:`
 
     2. You can also use our nifty capability to :ref:`automatically generate <autogen_configuration>` ``rsmtool`` configuration files rather than creating them manually.
 
+    3. To view the tutorial output on a `Weights & Biases <https://wandb.ai/>`__ project, use the file
+       `config_rsmtool_with_wandb.json` in the same tutorial folder. Make sure to edit the
+       :ref:`W&B entity <wandb_entity>` field to work with your Weights & Biases credentials.
+
 Run the experiment
 ^^^^^^^^^^^^^^^^^^
 Now that we have our features in right format and our configuration file in ``.json`` format, we can use the :ref:`rsmtool <usage_rsmtool>` command-line script to run our modeling experiment.
```
mulhod, line 89: Sorry, I know this isn't a change you made. Just spotted a minor issue.
```suggestion
Now that we have our features in the right format and our configuration file in ``.json`` format, we can use the :ref:`rsmtool <usage_rsmtool>` command-line script to run our modeling experiment.
```

doc/usage_rsmtool.rst
```diff
@@ -43,6 +43,8 @@ Output
...
+the specified Weights and Biases project.
```
mulhod, line 47:
```suggestion
the specified Weights & Biases project.
```

requirements.txt
```diff
@@ -17,5 +17,6 @@ sphinx<6.0
 sphinx_rtd_theme
 statsmodels
 tqdm
+wandb
```
mulhod, line 20: This is debatable, but I wonder if we really want to add `wandb` as a dependency. (And maybe this is just in `requirements.txt`, but not an actual dependency that would be in the `Conda`/`pip` packages, so correct me if I'm mistaken.) I thought this would an optional dependency, basically.

rsmtool/configuration_parser.py
```diff
@@ -1058,7 +1050,18 @@ def validate_config(cls, config, context="rsmtool"):
                     group: new_config["min_n_per_group"] for group in new_config["subgroups"]
                 }
 
-        # 15. Clean up config dict to keep only context-specific fields
+        # 15. Check that if the user enables logging to W&B, they also
+        # specified wandb project and entity.
+        if new_config["use_wandb"]:
```
mulhod, line 1055: While I don't mind having an extra key to make the use of W&B explicit, can't we just infer whether W&B is to be used or not based on the presence of `"wandb_project"` and `"wandb_entity"`?

rsmtool/rsmtool.py
```diff
@@ -15,6 +15,8 @@
 from os import listdir, makedirs
 from os.path import abspath, exists, join
 
+import wandb
```
mulhod, line 18: See my comment about optionality of `wandb`.

rsmtool/rsmtool.py
```diff
@@ -109,6 +111,15 @@ def run_experiment(config_file_or_obj_or_dict, output_dir, overwrite_output=Fals
     logger.info("Saving configuration file.")
     configuration.save(output_dir)
 
+    # If wandb logging is enabled, start a wandb run and log configuration
+    use_wandb = configuration.get("use_wandb", False)
+    wandb_run = None
+    if use_wandb:
+        wandb_project = configuration.get("wandb_project")
```
mulhod, line 118: I think these `.get`s should just be direct accesses. We probably don't want them ending up as `None`s. But also the validation checks for the presence of both fields, so we currently wouldn't get here without having them both be present.

rsmtool/writer.py
```diff
@@ -7,10 +7,13 @@
 
 :organization: ETS
 """
-
 from os import makedirs
 from os.path import join
 
+import wandb
```
mulhod, line 13: See comment about optionality.

mulhod (changes requested):

desilinguist (changes requested):

requirements.txt
desilinguist: If we make it an optional dependency, then we have to handle it like we handle "gridmap" in SKLL and check for `ImportError` etc.

desilinguist: Looking at their dependencies, I don' think it's that bad. Personally, I am okay with adding it as a straight-up dependency given that it will make our work at ETS easier for sure. And, like you said, it doesn't seem to be a thing that would bother people outside ETS anyway.

rsmtool/configuration_parser.py
mulhod: Yeah, that's true. Could support faster debugging/development.

desilinguist:
I think we will need to add tests, @tamarl08. That's not a minor change in coverage.

tamarl08:
Thanks for the review @desilinguist @mulhod! I addressed all your comments, leaving the dependency as it is for now.

desilinguist:
Okay, just tested this. Couple of things that we should add to the documentation:

1. I hadn't set up my W&B API key and so I just got this weird menu asking me to choose one of three options (create account, use existing account etc.). We should add a link to [Step 2](https://docs.wandb.ai/quickstart#2-log-in-to-wb) and tell folks that they should have completed this step before they start using W&B. 

2. If you enable W&B logging, the rsmtool run will take significantly longer due to the network traffic being sent to W&B. We should mention this.

Both of these can go under an "Important" note in the `use_wandb` section of the documentation.

doc/config_rsmtool.rst.inc
```diff
@@ -369,3 +369,23 @@ If set to ``true``, use the ``min`` and ``max`` columns specified in the ``featu
...
+use_wandb *(Optional)*
+""""""""""""""""""""""
+If set to ``true``, the generated reports and all intermediate tables will be logged to `Weights & Biases <https://wandb.ai/>`__.
+The Weights & Biases :ref:`entity<_wandb_entity>` and :ref:`project name<_wandb_project_name>` should be specified in the appropriate configuration fields. Defaults to ``false``.
+
+.. _wandb_project:
+
+wandb_project *(Optional)*
+""""""""""""""""""""""""""
+The Weights & Biases project name if logging to Weights & Biases is enabled.
```
desilinguist, line 384:
```suggestion
The Weights & Biases project name if logging to Weights & Biases is enabled. If a project by this name does not already exist, it will be created. 
```

desilinguist (approved):
LGTM, thanks!! 🎉

PR #621, opened 2023-07-21
Add wandb logging in rsmeval

tamarl08:
- Refactored wandb logging to a separate util file and added tests
- Added wandb logging to rsmeval


rsmtool/utils/wandb_logging.py
```diff
@@ -0,0 +1,72 @@
+"""
+Utility classes and functions for logging to Weights & Biases.
+
+:author: Tamar Lavee (tlavee@ets.org)
+"""
+import wandb
+
+EXCLUDE_WANDB_LOG = ["confMatrix"]
```
desilinguist, line 8: Are we excluding this because it's already in the report?

tests/data/experiments/dummyregressor/output/DummyRegressor_confMatrix.csv
desilinguist, line 1: Why are files being deleted?

desilinguist (changes requested):
I really like the cleaner wand setup now! Some minor comments.

rsmtool/utils/wandb_logging.py
tamarl08: Wandb throws an exception when logging it, I think it's because it has an empty cell at 0,0. I will look into it this week but didn't want to waste too much time on it before I finish the integration.

desilinguist (approved):

PR #624, opened 2023-07-27
Remove unnecessary warnings in HTML reports.

### LINKED ISSUE #618: Fix matplotlib warnings in reports
Seeing weird warnings before every figure in the report that say "The figure layout has changed to tight". We should figure out what the issue is and fix it.

desilinguist:
- Update the header notebooks for all tools to exclude the matplotlib and tqdm warnings that were being shown in all the reports.
- Replace unnecessary use of `resetwarnings()` with the `catch_warnings()` context manager in `fairness_utils.py`.  This is necessary to ensure that the filtering of warnings in notebooks works as expected.

To test this, please check out the branch, run an experiments for each of the command-line tools, and confirm that no warnings are shown in the reports.

This PR closes #618.


tamarl08 (approved):
Tested all the reports and no warnings!

damien2012eng (approved):

PR #627, opened 2023-07-28
Draft: Add wandb to rsmexplain and rsmxval

tamarl08:
Added the option to pass a wandb_run object to all modules - that allows rsmxval to use a single object and also will allow passing it from external runs which can be useful.


rsmtool/rsmexplain.py
```diff
@@ -399,10 +411,10 @@ def generate_explanation(
     )
 
     # generate the HTML report
-    generate_report(explanation, output_dir, ids, configuration, logger)
+    generate_report(explanation, output_dir, ids, configuration, logger, wandb_run=wandb_run)
 
 
-def generate_report(explanation, output_dir, ids, configuration, logger=None):
+def generate_report(explanation, output_dir, ids, configuration, logger=None, wandb_run=None):
```
mulhod, line 417: Add `wandb_run` to docstring.

desilinguist (approved):
I tried out the `rsmexplain` and `rsmxval` tutorial and they both work great. The `rsmxval` one was a bit on the slow side but I figured that was because there was a lot to log. 

It would be nice to log more intermediate files from both tools but that can be an additional issue.

mulhod (approved):

PR #632, opened 2023-07-31
Enhancements to fast_predict

### LINKED ISSUE #631: Enhance fast_predict by requiring fewer arguments (just the model)
It would be nice if we could make `feature_info`, `trim_min`, `trim_max`, and `trim_tolerance` attributes of the `Modeler` object. We can still expose these things in the `fast_predict` function. We could use the values if they are provided and not `None`. Otherwise, we check for the presence of these attributes and, if specified, use them instead. For `feature_info`, it must be able to be found as an attribute if `None`/not specified in the call to `fast_predict`. For `trim_min`/`trim_max`, if it's both `None`/unspecified and cannot be found as an attribute, the default value `None` will be used for (no trimming). For `trim_tolerance`, if it's both `None`/unspecified and cannot be found as an attribute, the default value of `0.4998` will be used.

This should make it possible to call `fast_predict` with only the data and a `Modeler` object.

mulhod:
Save RSMTool modeler in place of SKLL learner and make attributes like `feature_info`, `train_predictions_mean,` `trim_min`, etc., available in the modeler instance; modify `fast_predict` to be able to utilize these attributes in a backwards-compatible way so that fewer arguments can be passed to `fast_predict`.

~~This is a draft at the moment! No tests have been added quite yet. I would like to get some feedback on the direction here. Here are some points that I anticipate some discussion around:~~

1. `fast_predict` can now be called with just the features and the modeler instance. Parameters like `df_feature_info`, `trim_min`, and `h1_mean`, if left unspecified, will be retrieved from the modeler (provided it was trained with this or a future version of RSMTool).
2. Scaling and trimming in `fast_predict` can now be turned on/off via the `scale`/`trim` parameters, both of which are set to `False` by default. This is partially to allow for backwards-compatibility: If scaling-related parameters are not specified (currently), no scaling is done. But the attributes will always be there in the future (in the modeler object). Thus, this mechanism can be used to explicitly turn off trimming or scaling.
3. The change is backwards-compatible! Old RSMTool models (serialized SKLL learners) can still be loaded. Use of `fast_predict`, furthermore, is backwards-compatible. All of the attributes of the modeler that would be used for things like trimming and scaling can be explicitly passed in in the same way they are now in the call to `fast_predict`. Thus, old models and old `fast_predict` calls should work without requiring any change at all.
4. This change results in not saving SKLL models directly. Instead RSMTool modeler instances would be saved. These objects contain the SKLL model under the `learner` attribute, so we wouldn't really be losing much since it could always be referenced or saved by itself later.

Closes #631


mulhod:
ICYI, here's what I was doing to test that this works (at least for loading RSMTool modeler objects, so I haven't tested the backwards compatibility yet):

```shell
❯ ../../../env/bin/rsmtool config_rsmtool.json -f
Output directory: /Users/mmulholland/Documents/rsmtool/rsmtool/examples/rsmtool
WARNING: /Users/mmulholland/Documents/rsmtool/rsmtool/examples/rsmtool already contains a non-empty 'output' directory. Th
e generated report might contain unexpected information from a previous experiment.
Saving configuration file.
Reading in all data from files.
Preprocessing all features.
Pre-processing training and test set features
Saving training and test set data to disk.
Training LinearRegression model.
Running analyses on training set.
Generating training and test set predictions.
Processing train set predictions.
Processing test set predictions.
Scaling the coefficients and saving them to disk
Running prediction analyses.
Starting report generation.
Merging sections
Exporting HTML
WARNING: Alternative text is missing on 8 image(s).
❯ ../../../env/bin/ipython
Python 3.11.4 | packaged by conda-forge | (main, Jun 10 2023, 18:08:41) [Clang 15.0.7 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.14.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import rsmtool

In [2]: m = rsmtool.rsmtool.Modeler.load_from_file("output/ASAP2.model")

In [3]: rsmtool.rsmpredict.fast_predict({"DISCOURSE": 4.93806460126142, "ORGANIZATION": -0.0846667513334603, "GRAMMAR": -0.316793975540994, "MECHANICS": 4.65591397849462}, m)
Out[3]:
{'raw': 3.4671587960793446,
 'scale': 3.4876896893346814,
 'raw_trim': 3.4671587960793446,
 'raw_trim_round': 3,
 'scale_trim': 3.4876896893346814,
 'scale_trim_round': 3}
```

rsmtool/rsmtool.py
```diff
@@ -260,6 +262,16 @@ def run_experiment(config_file_or_obj_or_dict, output_dir, overwrite_output=Fals
...
+    pred_config_dict = pred_config.to_dict()
+    for key, attr_name in [
+        ("train_predictions_mean", "train_predictions_mean"),
+        ("train_predictions_sd", "train_predictions_sd"),
+        ("human_labels_mean", "h1_mean"),
+        ("human_labels_sd", "h1_sd"),
+    ]:
+        setattr(modeler, attr_name, pred_config_dict[key])
+    with open(join(csvdir, f"{configuration['experiment_id']}.model"), "wb") as model_file:
+        joblib.dump(modeler, model_file)
```
desilinguist, line 274: I think instead of doing this here, it might be better to add `save()` method to `Modeler` that encapsulates this and then call `modeler.save(model_file)` here.

desilinguist:
Hmm, One thing that immediately occurs to me is this – saving a modeler instead of a SKLL model – would be a completely backwards-incompatible change. 

At the very least, we would need to modify how `rsmpredict` and `rsmexplain` work at the very least since they need access to the model. Also, we would need to change notebook sections that try to inspect the model object. Probably lots of stuff in the documentation too.

mulhod:
Oh, yeah, `rsmpredict` is being modified now. I started with `fast_predict` right now as a POC. But the idea would be to have logic that tries to load a Learner object and, if that fails, load a Modeler object (or vice versa).

mulhod:
I'm going to push a change that should make that aspect a little bit clearer.

rsmtool/rsmtool.py
mulhod: Yeah, I'll do that.

mulhod:
@desilinguist I think this is ready for the part you were going to handle (converting unit test models).

tests/test_experiment_rsmpredict.py
```diff
@@ -358,3 +365,113 @@ def test_fast_predict_non_numeric(self):
         }
         with self.assertRaises(ValueError):
             _ = fast_predict(input_features, modeler, df_feature_info)
+
+    def test_fast_predict_no_feature_info(self):
+        """
+        Check that ``fast_predict()`` raises an error when
+        ``df_feature_info=None`` and the ``feature_info`` attribute
+        cannot be found.
```
desilinguist, line 385: Usually you want docstrings to be a one-line summary then a blank line and then a description.

tests/test_experiment_rsmpredict.py
```diff
@@ -358,3 +365,113 @@ def test_fast_predict_non_numeric(self):
...
+            "FEATURE7": 12,
+            "FEATURE8": -7000,
+        }
+        with self.assertRaises(ValueError):
+            _ = fast_predict(input_features, modeler)
+
+    def test_fast_predict_scaling_params_but_scale_false(self):
+        """
+        Check that ``fast_predict()`` raises an error when
+        scaling-related parameters are set but ``scale=False``.
```
desilinguist, line 418: docstring

tests/test_modeler.py
```diff
@@ -47,3 +49,24 @@ def test_expected_scores_no_min_max(self):
             ValueError, r"Must specify 'min_score' and 'max_score' for expected scores."
         ):
             self.modeler.predict(df, predict_expected=True)
+
+    def test_save_and_load(self):
```
desilinguist, line 53: Missing docstring.

tests/test_modeler.py
```diff
@@ -47,3 +49,24 @@ def test_expected_scores_no_min_max(self):
...
+        Modeler.load_from_file(model_file.name)
+        os.unlink(model_file.name)
+
+    def test_load_from_skll_model_file(self):
+        """
+        Meant to test for backward compatibility. Saved models were
+        formerly bare SKLL learner objects. This can be emulated by
+        saving just the learner attribute within the modeler object
+        and then trying to load that in the same way modeler objects
+        would be loaded.
```
desilinguist, line 66: Docstring.

desilinguist (commented):
Thanks for working on this @mulhod! I had some suggestions/comments, mostly minor.

rsmtool/rsmpredict.py
```diff
@@ -70,13 +70,13 @@ def fast_predict(
         features on which the model was trained and the values should
         be the *raw* feature values.
     modeler : rsmtool.modeler.Modeler object
-        The RSMTool modeler object using which the predictions
-        are to be generated. This object should be created from the already
-        existing model file in the "output" directory of the previously run
-        RSMTool experiment.
+        The RSMTool modeler object using which the predictions are to be
```
blongwill, line 73:
```suggestion
        The RSMTool modeler object from which the predictions are to be
```

rsmtool/rsmpredict.py
```diff
@@ -98,63 +98,63 @@ def fast_predict(
     trim : bool
         Whether to trim the predictions. If ``True``, ``trim_min`` and
         ``trim_max`` must be specified or be available as attributes of
-        the modeler object.
+        the ``modeler``.
```
blongwill, line 101: If load_from_file() and init() both have the object name capitalized in their docstring, perhaps keep that convention for consistency:
```suggestion
        the ``Modeler``.
```

rsmtool/modeler.py
```diff
@@ -34,14 +35,37 @@ class Modeler:
...
+        joblib.dump(self, model_path)
+
     @classmethod
     def load_from_file(cls, model_path):
         """
-        Load a modeler object from a file on disk.
+        Load a ``Modeler`` object from a file on disk.
 
         Parameters
         ----------
```
blongwill, line 71: Shouldn't the mention of modeler in model_path description get the backticks and capitalization also?

blongwill: Also the same thing under returns:  A ``Modeler`` instance.

rsmtool/rsmpredict.py
```diff
@@ -68,11 +70,14 @@ def fast_predict(
...
-        The RSMTool modeler object using which the predictions
-        are to be generated. This object should be created from the already
-        existing model file in the "output" directory of the previously run
-        RSMTool experiment.
-    df_feature_info : pandas DataFrame
+        The RSMTool modeler object using which the predictions are to be
+        generated. This object should be created from the already existing
+        model file in the "output" directory of the previously run RSMTool
+        experiment.
+    df_feature_info : pandas DataFrame, optional
```
blongwill, line 77: This is such an elegant solution to avoid the clunkiness of storing separate feature info files alongside the model. Great idea!

rsmtool/rsmpredict.py
```diff
@@ -164,24 +205,53 @@ def fast_predict(
...
+        if trim_max is None:
+            trim_max = modeler.trim_max
         for column in df_predictions.columns:
             df_predictions[f"{column}_trim"] = preprocessor.trim(
                 df_predictions[column], trim_min, trim_max, trim_tolerance
             )
             df_predictions[f"{column}_trim_round"] = np.rint(
                 df_predictions[f"{column}_trim"]
             ).astype("int64")
+    elif any(arg is not None for arg in [trim_tolerance, trim_min, trim_max]):
```
blongwill, line 251: Not to go into a validation rabbit hole, but perhaps we should consider adding validation for the opposite case as well? If backward compatibility is going to be possible, someone might specify trim=True  expecting one of the older models to provide min/max values when it can't. Instead of letting it give the unsupported operand error in preprocessor.trim, it might be good to give a more informative error message before it?

rsmtool/rsmpredict.py
```diff
@@ -88,41 +93,68 @@ def fast_predict(
 
         This dataframe should be read from the "feature.csv" file under the
         "output" directory of the previously run RSMTool experiment.
+
+        Defaults to ``None``.
+    trim : bool
```
blongwill, line 98:
```suggestion
    trim : bool, optional
```

rsmtool/rsmpredict.py
```diff
@@ -88,41 +93,68 @@ def fast_predict(
...
+        Whether to trim the predictions. If ``True``, ``trim_min`` and
+        ``trim_max`` must be specified or be available as attributes of
+        the ``modeler``.
+        Defaults to ``False``.
     trim_min : int, optional
         The lowest possible integer score that the machine should predict.
-        If ``None``, raw predictions will not be trimmed.
+        If ``None``, this function will try to extract this value from
+        ``modeler``. If ``None`` and no such attribute exists, raw
```
blongwill, line 106: _"If ``None`` and no such attribute exists, raw predictions will not be trimmed."_

if `trim` is set to True, then won't this scenario result in a TypeError [at this part of preprocessor.trim](https://github.com/EducationalTestingService/rsmtool/blob/4c5cab6b20f845426f34998dc652241a3fe2ebd5/rsmtool/preprocessor.py#L413-L414)?

mulhod: It would have led to an AttributeError, I believe. However, I have converted it to a ValueError now with a helpful message. Of course, if the attribute or what you pass in is not a float, then, yes, I think a TypeError would be thrown. I'm not sure it's worth it to validate to that level.

rsmtool/rsmpredict.py
mulhod: I have changed this. However, in my opinion I think it should be reserved perhaps for parameters that could be ``None`` rather than those that simply have a default. In typing, we use `Optional` to mean some value or `None`. I like to stick to that meaning. But I think the meaning in this codebase is a bit broader.

rsmtool/rsmpredict.py
mulhod: I think it would have resulted in an AttributeError (now a ValueError, though). I believe it should be somewhat easy to understand what's going on either way.

mulhod:
Thanks, @blongwill! I believe that I have addressed your concerns. Please take another look!

rsmtool/rsmpredict.py
desilinguist: Yeah, that's true. When (If?) we go through the process of adding type hints to everything, we will resolve this inconsistency.

rsmtool/modeler.py
mulhod: I have gone through a lot of the documentation and tried to make things more consistent. It's not always the case, though, that I'm referring to `Modeler` or a `modeler` (that is, if that's the name of the parameter itself, in which case I'm not talking about the class). I think it should be relatively clear now.

desilinguist:
@mulhod  I just updated all the relevant model files. I just realized we haven't yet looked at the documentation. I think there may be few places where the `.model` file may be described (the outputs generated by rsmtool). Can you please take a look and update all the relevant places?

mulhod:
@desilinguist I have done some grepping around in the documentation. I found one instance so far that needed to be changed. I'll continue digging a bit, though. My first iteration was looking specifically for pieces that touched on literally a `.model` file output. But, perhaps search for `SKLL`, etc., would be a good idea too.

doc/intermediate_files_rsmtool.rst.inc
```diff
@@ -178,7 +178,7 @@ Model information
 
 - ``model_fit``: R squared and adjusted R squared computed on the training set. Note that these values are always computed on raw predictions without any trimming or rounding.
 
-- ``.model``: the serialized SKLL ``Learner`` object containing the fitted model (before scaling the coeffcients).
+- ``.model``: the serialized RSMTool ``Modeler`` object containing the SKLL ``Learner`` object with the fitted model (before scaling the coeffcients).
```
desilinguist, line 181:
```suggestion
- ``.model``: the serialized RSMTool ``Modeler`` object containing the fitted SKLL ``Learner`` object  (before scaling the coeffcients).
```

desilinguist (approved):
LGTM! 🎉

rsmtool/rsmpredict.py
blongwill: @mulhod It doesn't give an error like this when trying to add the max_trim to the tolerance?
TypeError: unsupported operand type(s) for +: 'NoneType' and 'float'

blongwill:
> Thanks, @blongwill! I believe that I have addressed your concerns. Please take another look!

Looks great Matt! Thanks for making those updates!

rsmtool/rsmpredict.py
mulhod: This merged now, but I wanted to show you what happens when, say, `trim_max` doesn't exist as an attribute. It's caught before it ever gets to the point where `trim_max` would be added to `trim_tolerance`.

```shell
❯ ../env/bin/ipython
Python 3.11.4 | packaged by conda-forge | (main, Jun 10 2023, 18:08:41) [Clang 15.0.7 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.14.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from rsmtool.rsmpredict import fast_predict

In [2]: from rsmtool.modeler import Modeler

In [3]: m = Modeler.load_from_file("tests/data/experiments/lr-predict/existing_experiment/output/lr.model")

In [4]: features = {
   ...:             "FEATURE1": 6.0,
   ...:             "FEATURE2": 1.0,
   ...:             "FEATURE3": -0.2,
   ...:             "FEATURE4": 0,
   ...:             "FEATURE5": -0.1,
   ...:             "FEATURE6": 5.0,
   ...:             "FEATURE7": 12,
   ...:             "FEATURE8": -7000,
   ...:         }

In [5]: delattr(m, "trim_max")

In [6]: fast_predict(features, m, trim=True)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[6], line 1
----> 1 fast_predict(features, m, trim=True)

File ~/Documents/rsmtool/rsmtool/rsmtool/rsmpredict.py:258, in fast_predict(input_features, modeler, df_feature_info, trim, trim_min, trim_max, trim_tolerance, scale, train_predictions_mean, train_predictions_sd, h1_mean, h1_sd, logger)
    256         trim_max = modeler.trim_max
    257 except AttributeError:
--> 258     raise ValueError(
    259         "The trim_min/trim_max modeler attributes must exist when trim=True "
    260         "and these values are not explicitly specified"
    261     ) from None
    262 for column in df_predictions.columns:
    263     df_predictions[f"{column}_trim"] = preprocessor.trim(
    264         df_predictions[column], trim_min, trim_max, trim_tolerance
    265     )

ValueError: The trim_min/trim_max modeler attributes must exist when trim=True and these values are not explicitly specified
```

Now, If we specifically pass in a value for `trim_max`, in which case it will not check for the attribute, and that value is not a number of some kind, then it will result in a `TypeError`. But, this is expected since the value passed in is not a valid `trim_max`. This could be caught and validated, but 1) it's a strange thing to do and 2) the error message is not unclear. This is why I don't think it's worth it to validate. It would still be an error, so it's not like it would silently fail.

```shell
In [7]: fast_predict(features, m, trim=True, trim_max="not_a_number")
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[10], line 1
----> 1 fast_predict(features, m, trim=True, trim_max="not_a_number")

File ~/Documents/rsmtool/rsmtool/rsmtool/rsmpredict.py:263, in fast_predict(input_features, modeler, df_feature_info, trim, trim_min, trim_max, trim_tolerance, scale, train_predictions_mean, train_predictions_sd, h1_mean, h1_sd, logger)
    258         raise ValueError(
    259             "The trim_min/trim_max modeler attributes must exist when trim=True "
    260             "and these values are not explicitly specified"
    261         ) from None
    262     for column in df_predictions.columns:
--> 263         df_predictions[f"{column}_trim"] = preprocessor.trim(
    264             df_predictions[column], trim_min, trim_max, trim_tolerance
    265         )
    266         df_predictions[f"{column}_trim_round"] = np.rint(
    267             df_predictions[f"{column}_trim"]
    268         ).astype("int64")
    269 elif any(arg is not None for arg in [trim_tolerance, trim_min, trim_max]):

File ~/Documents/rsmtool/rsmtool/rsmtool/preprocessor.py:413, in FeaturePreprocessor.trim(self, values, trim_min, trim_max, tolerance)
    410 if isinstance(values, list):
    411     values = np.array(values)
--> 413 new_max = trim_max + tolerance
    414 new_min = trim_min - tolerance
    415 trimmed_values = values.copy()

TypeError: can only concatenate str (not "float") to str
```

blongwill: @mulhod But in normal usage the trim_min/max attributes are set to None when the model is instantiated, right?. I think you're only getting the attribute error by artificially deleting the model attribute at In [5] and then later trying to set the nonexistent variable. With the variable in existence as None I think it would fall into that second case you demonstrated.

blongwill: In the standard usage line 256 (i.e., trim_max = modeler.trim_max) would just be setting trim_max to None, right? Am I  still missing something here? Either way, thanks for taking the time to discuss this further even after it's been merged!

mulhod: Perhaps I don't understand what you mean. Could you demonstrate it with an example?

blongwill: I don't see any reason in the code that the modeler.trim_max instance wouldn't exist. Even if you're working with a skll model, load_from_learner() still instantiates a new Modeler instance and just sets the learner attribute. So the other class attributes should still exists and are set to None

mulhod: You're right that instantiating a modeler instance will have those attributes set to `None`, but those instances are only saved as part of running `rsmtool`, in which case the attributes would be there. Basically, we don't typically instantiate an instance anywhere outside of `rsmtool`, so modeler objects are not saved by anything other than `rsmtool`. The only case where the `TypeError` would occur is if you instantiated a modeler manually and then just saved it without really doing anything else. It's the kind of edge case that I'm not sure we need to consider right now. Even if you were doing something like this, you would have to have good knowledge about the internal workings of these things, so it probably wouldn't need to be caught.

blongwill: > The only case where the TypeError would occur is if you instantiated a modeler manually and then just saved it without really doing anything else

A much more common occurrence would be using  load_from_learner() with a model that was built/saved with the previous version of RSMTool and the result would be the same, wouldn't it? Either way I agree that this isn't worth spending any more time on since it is just an edge case. 

Although, @mulhod Do you think there could be any argument for adding a model version attribute to the newer rsmtool model objects? Then when load_from_learner() is used, it could automatically display the version of the model to help distinguish between older models and newer?

mulhod: @blongwill I hope this answers your question. Below I load in an old model and try to make a prediction with it. The attribute is not None, it simply doesn't exist. It throws the expected error.

```shell
# Check out old model
❯ git checkout v9.1.1 tests/data/experiments/lr-predict/existing_experiment/output/lr.model
❯ ../env/bin/ipython
Python 3.11.4 | packaged by conda-forge | (main, Jun 10 2023, 18:08:41) [Clang 15.0.7 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.14.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from rsmtool.modeler import Modeler

In [2]: from rsmtool.rsmpredict import fast_predict

In [3]: m = Modeler.load_from_file("tests/data/experiments/lr-predict/existing_experiment/output/lr.model")
Trying to unpickle estimator StandardScaler from version 1.2.0 when using version 1.3.0. This might lead to breaking code or invalid results. Use at your own risk. For more info please refer to:
https://scikit-learn.org/stable/model_persistence.html#security-maintainability-limitations
Trying to unpickle estimator LinearRegression from version 1.2.0 when using version 1.3.0. This might lead to breaking code or invalid results. Use at your own risk. For more info please refer to:
https://scikit-learn.org/stable/model_persistence.html#security-maintainability-limitations

In [4]: m.feature_info
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[4], line 1
----> 1 m.feature_info

AttributeError: 'Modeler' object has no attribute 'feature_info'

In [5]: features = {
   ...:             "FEATURE1": 6.0,
   ...:             "FEATURE2": 1.0,
   ...:             "FEATURE3": -0.2,
   ...:             "FEATURE4": 0,
   ...:             "FEATURE5": -0.1,
   ...:             "FEATURE6": 5.0,
   ...:             "FEATURE7": 12,
   ...:             "FEATURE8": -7000,
   ...:         }

In [6]: fast_predict(features, m)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[6], line 1
----> 1 fast_predict(features, m)

File ~/Documents/rsmtool/rsmtool/rsmtool/rsmpredict.py:197, in fast_predict(input_features, modeler, df_feature_info, trim, trim_min, trim_max, trim_tolerance, scale, train_predictions_mean, train_predictions_sd, h1_mean, h1_sd, logger)
    195         df_feature_info = modeler.feature_info
    196     except AttributeError:
--> 197         raise ValueError(
    198             "df_feature_info must be specified if it cannot be found as an "
    199             "attribute in the modeler object"
    200         ) from None
    202 # preprocess the input features so that they match what the model expects
    203 try:

ValueError: df_feature_info must be specified if it cannot be found as an attribute in the modeler object
```

Can you tell me what about this might be unexpected or problematic?

mulhod: Ok, I think I may understand what you're getting at now. When loading from a bare SKLL model file, the `trim_min`, etc., attributes will actually exist since they are set to `None`. That is, they won't simply be missing, which my tests were assuming. They will be there and have a value of `None`. So, this would be a problem. It's not a very big problem because they would encounter an exception eventually (a `TypeError`, as you mention). But, it's nice to catch this earlier perhaps so that it's clear what the problem is. I will make a merge request to show what I mean.

PR #633, opened 2023-07-31
Include system information in RSMExplain reports

### LINKED ISSUE #625: RSMExplain reports should also contain system information
The rsmexplain reports are missing the system information with the rsmtool version, the operating system version, and the package versions. We should include this just like in the other reports.

desilinguist:
- Fix import sorting in RSMExplain's `header.ipynb` and include additional imports needed for system information.
- Add `sysinfo.ipynb` notebook for RSMExplain. 
- Add "`sysinfo.ipynb`" in default RSMExplain sections in `reporter.py`.
- Update RSMExplain documentation to include `general_sections` option which was missing. 
- Update expected output for auto-generated configuration functions to include "sysinfo". 

To test, run the `rsmexplain` tutorial and confirm that the system information section is included in the report.

Closes #625.


doc/config_rsmexplain.rst.inc
```diff
@@ -39,16 +39,34 @@ description *(Optional)*
...
 
+num_features_to_display *(Optional)*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Number of top features that should be displayed in ``rsmexplain`` plots. Defaults to 15.
+
+.. _rsmexplain_sample_ids:
+
+sample_ids *(Optional)*
+~~~~~~~~~~~~~~~~~~~~~~~
+If we want to explain a specific set of responses from the ``explain_data``, we can specify their IDs here as a comma-separated string. Note that the IDs must be values from the :ref:`id_column <rsmexplain_id_column>`. For example, if ``explain_data`` has IDs of the form ``"EXAMPLE_1"``, ``"EXAMPLE_2"`` etc. and we want to explain the fifth, tenth, and twelfth example, the value of this field must be set to ``"EXAMPLE_5, EXAMPLE_10, EXAMPLE_12"``. Defaults to ``None``.
```
mulhod, line 68:
```suggestion
If we want to explain a specific set of responses from the ``explain_data``, we can specify their IDs here as a comma-separated string. Note that the IDs must be values from the :ref:`id_column <rsmexplain_id_column>`. For example, if ``explain_data`` has IDs of the form ``"EXAMPLE_1"``, ``"EXAMPLE_2"``, etc., and we want to explain the fifth, tenth, and twelfth example, the value of this field must be set to ``"EXAMPLE_5, EXAMPLE_10, EXAMPLE_12"``. Defaults to ``None``.
```

mulhod (approved):

doc/config_rsmexplain.rst.inc
desilinguist: I am going to fix that in the next MR - don't want to run the build again for this tiny change :)

PR #641, opened 2023-08-08
Improvements to fast_predict validation/error handling

mulhod:
Deal with situation where trimming, scaling, etc., modeler attributes exist but are `None`, which would be the case when loading an older model (bare SKLL model).

Even if these attributes were `None` before, it would still have led to an error somewhere down the line (`TypeError` if `trim_min` and `trim_tolerance` were `None` and/or a mixture of `None` and `float`, for instance). This change should catch that and give a nicer error message.

### DIFF
--- a/rsmtool/rsmpredict.py
+++ b/rsmtool/rsmpredict.py
@@ -173,13 +173,17 @@ def fast_predict(
     Raises
     ------
     ValueError
-        If ``input_features`` contains any non-numeric features; if
-        trimming/scaling is turned on but related parameters are either
+        If ``input_features`` contains any non-numeric features
+    ValueError
+        If trimming/scaling is turned on but related parameters are either
         not specified or cannot be found as attributes in ``modeler``/have
-        a value of ``None``; if trimming/scaling-related parameters are
-        specified but trimming/scaling is turned off; if feature
-        information is either not specified or cannot be found as an
-        attribute in ``modeler``/has a value of ``None``.
+        a value of ``None``
+    ValueError
+        If trimming/scaling-related parameters are specified but
+        trimming/scaling is turned off
+    ValueError
+        If feature information is either not specified or cannot be found
+        as an attribute in ``modeler``/has a value of ``None``
     """
     # initialize a logger if none provided
     logger = logger if logger else logging.getLogger(__name__)
@@ -192,16 +196,17 @@ def fast_predict(
     df_input_features["spkitemid"] = "RESPONSE"
 
     feature_info_error_message = (
-        "df_feature_info must be specified if it cannot be found as an attribute "
-        "in the modeler object with a non-None value"
+        "'df_feature_info' must be specified if it not found as an attribute in the "
+        "modeler object with a value that is not ``None``"
     )
-    try:
-        if df_feature_info is None:
-            df_feature_info = modeler.feature_info
-            if df_feature_info is None:
-                raise ValueError(feature_info_error_message) from None
-    except AttributeError:
-        raise ValueError(feature_info_error_message) from None
+    if df_feature_info is None:
+        try:
+            stored_feature_info = modeler.feature_info
+            assert stored_feature_info is not None
+        except (AttributeError, AssertionError):
+            raise ValueError(feature_info_error_message) from None
+        else:
+            df_feature_info = stored_feature_info
 
     # preprocess the input features so that they match what the model expects
     try:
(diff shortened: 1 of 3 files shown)


rsmtool/rsmpredict.py
```diff
@@ -175,10 +175,11 @@ def fast_predict(
     ValueError
         If ``input_features`` contains any non-numeric features; if
         trimming/scaling is turned on but related parameters are either
```
desilinguist, line 177: Generally, we have a separate ValueError declaration for each reason. So,

```
ValueError
    If ``input_features`` contains any non-numeric features
ValueError
    if trimming/scaling is turned on but related parameters are either 
    not specified or cannot be found as attributes in ``modeler``
...
```

rsmtool/rsmpredict.py
```diff
@@ -190,14 +191,17 @@ def fast_predict(
     df_input_features = pd.DataFrame([input_features])
     df_input_features["spkitemid"] = "RESPONSE"
 
-    if df_feature_info is None:
-        try:
+    feature_info_error_message = (
+        "df_feature_info must be specified if it cannot be found as an attribute "
```
desilinguist, line 195:
```suggestion
        "'df_feature_info' must be specified if not found as an attribute "
```

rsmtool/rsmpredict.py
```diff
@@ -190,14 +191,17 @@ def fast_predict(
...
+        "in the modeler object with a non-None value"
```
desilinguist, line 196:
```suggestion
        "in the modeler object with a value that is not ``None``."
```

rsmtool/rsmpredict.py
```diff
@@ -190,14 +191,17 @@ def fast_predict(
...
             df_feature_info = modeler.feature_info
-        except AttributeError:
-            raise ValueError(
-                "df_feature_info must be specified if it cannot be found as an "
-                "attribute in the modeler object"
-            ) from None
+            if df_feature_info is None:
+                raise ValueError(feature_info_error_message) from None
+    except AttributeError:
+        raise ValueError(feature_info_error_message) from None
```
desilinguist, line 204: Can't we simplify this like so?


```python
if df_feature_info is None:
    try:
        stored_feature_info = modeler.feature_info
        assert df_feature_info is not None
    except AttributeError, AssertionError:
        raise ValueError(feature_info_error_message) from None
    else:
        df_feature_info = stored_feature_info
```

rsmtool/rsmpredict.py
```diff
@@ -243,22 +253,27 @@ def fast_predict(
 
     # trim both raw and scaled predictions if requested
     if trim:
+        trim_args_error_message = (
+            "The trim_min/trim_max modeler attributes must exist and not be None when "
+            "trim=True and these values are not explicitly specified"
+        )
```
desilinguist, line 259:
```suggestion
        trim_args_error_message = (
            "When 'trim' is set to ``True`` and no explicit values are provided, the 'trim_min' and 
            "'trim_max' modeler attributes must be present and not ``None``."
        )
```

desilinguist (changes requested):

desilinguist (approved):
LGTM! Thanks for incorporating my suggestions!

mulhod:
@blongwill Please take a look.

blongwill (commented):
Thanks Matt this looks great! I made some surface level comments but nothing big, take em or leave em!

blongwill (approved):

rsmtool/rsmpredict.py
```diff
@@ -243,22 +258,28 @@ def fast_predict(
...
+                if trim_tolerance is None:
+                    trim_tolerance = default_trim_tolerance
             except AttributeError:
                 trim_tolerance = default_trim_tolerance
         try:
             if trim_min is None:
                 trim_min = modeler.trim_min
             if trim_max is None:
                 trim_max = modeler.trim_max
+            if any(arg is None for arg in [trim_tolerance, trim_min, trim_max]):
```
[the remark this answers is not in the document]
mulhod: Actually, it's designed to not catch that case on purpose. The thinking was that they should be able to override values if they want.

tests/test_experiment_rsmpredict.py
```diff
@@ -408,6 +407,35 @@ def test_fast_predict_no_feature_info(self):
...
+            "experiments",
+            "lr-predict",
+            "existing_experiment",
+            "output",
+        )
+        model_file = join(existing_experiment_dir, "lr.model")
+
+        # initialize the modeler instance with a bare learner, in which
+        # case no feature_info attribute will exist
+        modeler = Modeler.load_from_learner(Modeler.load_from_file(model_file).learner)
```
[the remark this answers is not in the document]
mulhod: What I wanted to achieve with this line was to load from the learner, but in order to load from the bare learner I first have to load from the modeler and then get its learner attribute. Then, I can reload from that learner attribute as if I had loaded from a learner from the start. Technically, this is done only once. There are similar lines in 2 other locations, but in those cases the calls are not chained together like this (since I need to get some other info from the modeler before loading from the learner). I don't think it's worth it to abstract this out into a utility function. In fact, I think it would just add complexity.

rsmtool/rsmpredict.py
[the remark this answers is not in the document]
blongwill: This statement catches what I'm referring to. If they only specify trim_min but not trim max leaving it as None, the exception below would be triggered. This case is just not demonstrated in the unit tests I don't think

PR #642, opened 2023-08-14
Fix test dir usage in test_wandb

tamarl08:


tamarl08:
@desilinguist I also made ruff auto fix errors, because I just couldn't sort it out :) let me know if we shouldn't have it for some reasons.

desilinguist (approved):

PR #650, opened 2023-08-31
Allow prallelization of grid search

tamarl08:


desilinguist:
@tamarl08  you'll need to rebase.

mulhod (approved):

doc/config_rsmtool.rst.inc
```diff
@@ -374,6 +374,14 @@ If set to ``true``, use the ``min`` and ``max`` columns specified in the ``featu
 
     If ``use_truncation_thresholds`` is set, a ``features`` file _must_ be specified, and this file _must_ include ``min`` and ``max`` columns. If no ``feature`` file is specified or these columns are missing, an error will be raised.
 
+.. _grid_search_jobs:
+
+grid_search_jobs *(Optional)*
+""""""""""""""""""""""
+Number of folds to run in parallel when using skll grid search. If set to ``None``, the number of grid search folds will
+be used. Defaults to 1, that is, no parallelization. For very large data sets, using parallel jobs may significantly improve
```
mulhod, line 382:
```suggestion
Number of folds to run in parallel when using ``skll`` grid search. If set to ``None``, the number of grid search folds will default to 1 (no parallelization). For very large data sets, using parallel jobs may significantly improve
```

mulhod (changes requested):
I kind of wonder about exposing this in the config since it's a detail that doesn't necessarily have to do with input/experiment. I don't know where else it would be put, but it seems slightly out of place. I'm not sure there's a better way to do it, but my initial thought was that it could be exposed by an environment variable. But I do see the benefit. I wonder if there's a possibility -- if it's not too complicated -- to also have this be controlled by an environment variable, though.

tamarl08:
@mulhod I see your point. I also don't like the fact that it defaults to 1, but can get a value of None to enable parallelization.

I didn't find anything else that we control using an environment variable, do we do that?

Another option I thought of, is adding a general option of sending extra skll args dict in the call to `run_experiment`. It will also allow disabling the grid search and specifying fixed parameters, and possibly many other tuning options, without making rsmtool config look like skll config.

desilinguist:
Honestly, I am not that bothered by the existence of this parameter since it's meant for advanced users who will probably already be familiar with the same parameter in SKLL. 

am not sure I see the benefits of using an environment variable over a configuration setting since then users have to think about two places rather than one for settings.

I think the idea of being able to pass a dictionary is great for advanced users but we should make that a longer term goal rather than for this PR.

mulhod:
I think what I was perhaps envisioning was something like:

Get environment variable "SKLL_GRID_JOBS" if it exists; if not, use "grid_jobs" from config; if neither, use `1`. This way, it could be passed in either way. But, it does make it more complicated.

doc/config_rsmtool.rst.inc
```diff
@@ -374,6 +374,14 @@ If set to ``true``, use the ``min`` and ``max`` columns specified in the ``featu
 
     If ``use_truncation_thresholds`` is set, a ``features`` file _must_ be specified, and this file _must_ include ``min`` and ``max`` columns. If no ``feature`` file is specified or these columns are missing, an error will be raised.
 
+.. _grid_search_jobs:
+
+grid_search_jobs *(Optional)*
+""""""""""""""""""""""
```
mulhod, line 380:
```suggestion
"""""""""""""""""""""""""""""
```

doc/config_rsmtool.rst.inc
```diff
@@ -374,6 +374,13 @@ If set to ``true``, use the ``min`` and ``max`` columns specified in the ``featu
 
     If ``use_truncation_thresholds`` is set, a ``features`` file _must_ be specified, and this file _must_ include ``min`` and ``max`` columns. If no ``feature`` file is specified or these columns are missing, an error will be raised.
 
+.. _grid_search_jobs:
+
+grid_search_jobs *(Optional)*
+"""""""""""""""""""""""""""""
+Number of folds to run in parallel when using skll grid search. Defaults to 1 (no parallelization). For very large data
+sets, using parallel jobs may significantly improve running time.
```
desilinguist, line 382:
```suggestion
sets, setting this option to a higher number (e.g., number of CPU cores) may significantly improve running time.
```

desilinguist:
> > @mulhod I see your point. I also don't like the fact that it defaults to 1, but can get a value of None to enable parallelization.
> > I didn't find anything else that we control using an environment variable, do we do that?
> > Another option I thought of, is adding a general option of sending extra skll args dict in the call to `run_experiment`. It will also allow disabling the grid search and specifying fixed parameters, and possibly many other tuning options, without making rsmtool config look like skll config.
> 
> By the way, this is how it works in `skll`.
> 
> https://github.com/EducationalTestingService/skll/blob/main/skll/utils/constants.py#L138 

Honestly, I've never really been a fan of that approach and I think most users of SKLL either don't know about it or use it. It also doesn't seem to be mentioned in the docs at all so it's more of a hidden setting.

desilinguist:
@tamarl08 since this new setting _only_ affects SKLL models, I think we should prefix it with `skll_` to call it `skll_grid_search_jobs` in keeping with other options like `skll_objective` that already exist. Thoughts?

desilinguist (changes requested):

desilinguist:
@tamarl08 did you already check that automatic and interactive config generation work with this new field?

tamarl08:
I added the field to the output and the tests pass, do we usually do manual testing as well?

desilinguist:
Yes, I usually like to test the config generation by running `rsmtool generate` and `rsmtool generate --interactive` just to be sure.

tamarl08:
> Yes, I usually like to test the config generation by running `rsmtool generate` and `rsmtool generate --interactive` just to be sure.

Done!

desilinguist (approved):
LGTM! Thank you!

mulhod (approved):

PR #654, opened 2023-09-08
Fix warnings in reports

### LINKED ISSUE #652: Fix warning in report: is_categorical_dtype is deprecated

### LINKED ISSUE #653: Fix warning in report: pkg_resources is deprecated as an API
See https://setuptools.pypa.io/en/latest/pkg_resources.html

tamarl08:
1. Suppress warning "is_categorical_dtype is deprecated..." coming from seaborn following the recent version of pandas. We should remove this when the next version of seaborn is released. 
2. use importlib_metadata to list packages instead of the deprecated pkg_resources.

closes #652, closes #653


desilinguist:
I am also seeing this warning on the console when running rsmtool:

```
Setting an item of incompatible dtype is deprecated and will raise in a future error of pandas. Value '-' has dtype incompatible with int64, please explicitly cast to a compatible dtype first.
```

desilinguist:
@tamarl08 we need another issue right?

desilinguist:
May be worth running all of the tools to check any other warnings in the reports and on the console, while we are at it?

tamarl08:
@desilinguist 
- I ran test with STRICT locally on this branch and it passes, does it check the console tools?
- I'll open an issue for followup on seaboarn
- I see these warning too. I was not 100% sure if they are new. should we resolve them before release too?

desilinguist:
1. No, that just checks warnings in reports.
2. Great!
3. Definitely new.

tamarl08:
@desilinguist I fixed the warning by setting these cells to None - this is better than -1 because then the cell in the table is empty which is nicer.
rsmexplain also has a new warning which is fixed in shap and will be included in their next version. I opened an issue for that too.

desilinguist:
Perfect! I think the `n_init` warning has been there for a while. Good to know that they are fixing it. 

Hmm, if I test with the `lr-exclude-flags-and-zeros` test, the table cell has a `NaN` in it, it's not empty.

CleanShot 2023-09-08 at 17 33 49@2x

desilinguist:
@tamarl08 speaking of warnings, I am seeing a couple of new ones when I run the `lr-exclude-flags-and-zeros`:

> DataFrame.applymap has been deprecated. Use DataFrame.map instead.
> DataFrame.applymap has been deprecated. Use DataFrame.map instead.

tamarl08:
> Perfect! I think the `n_init` warning has been there for a while. Good to know that they are fixing it.
> 
> Hmm, if I test with the `lr-exclude-flags-and-zeros` test, the table cell has a `NaN` in it, it's not empty. Also shouldn't the numbers all be integers instead of some being floats?
> CleanShot 2023-09-08 at 17 33 49@2x

Which report is this? I don't see this table.

Just found another warning in summarize:
```
Series.__getitem__ treating keys as positions is deprecated. In a future version, integer keys will always be treated as labels (consistent with DataFrame behavior). To access a value by position, use `ser.iloc[pos]`
```

desilinguist:
> > Perfect! I think the `n_init` warning has been there for a while. Good to know that they are fixing it.
> 
> > 
> 
> > Hmm, if I test with the `lr-exclude-flags-and-zeros` test, the table cell has a `NaN` in it, it's not empty. Also shouldn't the numbers all be integers instead of some being floats?
> 
> > CleanShot 2023-09-08 at 17 33 49@2x
> 
> 
> 
> Which report is this? I don't see this table.
> 

It's rsmtool using the test I mentioned. Run that manually.

PR #659, opened 2023-09-27
Add sections to W&B logging

tamarl08:
Metrics are logged to different sections according to the context (rsmtool, rsmeval etc.). That way they appear separately on the [W&B run](https://wandb.ai/etslabs/xval_new_3/runs/edfufp6d?workspace=user-tamarlv) and do not overwrite when more than one module log the same metrics.
All reports are logged to a `reports` section.

This PR also updates the RTD configuration to meet the new format of the `build` section.


doc/config_rsmtool.rst.inc
```diff
@@ -281,6 +281,13 @@ The values for a key can be string, integer, float, or boolean depending on what
 hyperparameter expects. Note that if this option is specified with the
 :ref:`built-in linear regression models <builtin_models>`, it will simply be ignored.
 
+.. _skll_grid_search_jobs:
+
+skll_grid_search_jobs *(Optional)*
+""""""""""""""""""""""""""""""""""
+Number of folds to run in parallel when using skll grid search. Defaults to 1 (no parallelization). For very large data
```
mulhod, line 288: Just a minor nitpick. It would be nice to distinguish the string `skill` with ticks to indicate it is the name of a library.

```suggestion
Number of folds to run in parallel when using ``skll`` grid search. Defaults to 1 (no parallelization). For very large data
```

mulhod:
I'd like to try these changes in a real scenario. Perhaps we can work together to do that, but it will take some work to do before this is a package that can be installed.

mulhod (approved):

desilinguist:
> I'd like to try these changes in a real scenario. Perhaps we can work together to do that, but it will take some work to do before this is a package that can be installed.

@tamarl08 let's build a conda package and send the tarball to @mulhod, he can install it locally and test?

rsmtool/analyzer.py
```diff
@@ -1820,7 +1824,11 @@ def run_prediction_analyses(self, data_container, configuration, wandb_run=None)
                 conf_matrix_h1h2, index=labels, columns=labels
             ).transpose()
             wandb.log_confusion_matrix(
-                wandb_run, human1_scores, human2_scores, "Human1-Human2 Confusion Matrix"
+                wandb_run,
+                human1_scores,
+                human2_scores,
```
mulhod, line 1829: For some reason, I'm hitting an error at this call to `log_confusion_matrix`. Adding `.values` makes it work, though. See the traceback below.

```suggestion
                human1_scores.values,
                human2_scores.values,
```

```
  File "/env/lib/python3.11/site-packages/pandas/core/indexes/base.py", line 3790, in get_loc
    return self._engine.get_loc(casted_key)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "index.pyx", line 152, in pandas._libs.index.IndexEngine.get_loc
  File "index.pyx", line 181, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/hashtable_class_helper.pxi", line 2606, in pandas._libs.hashtable.Int64HashTable.get_item
  File "pandas/_libs/hashtable_class_helper.pxi", line 2630, in pandas._libs.hashtable.Int64HashTable.get_item
KeyError: 3

The above exception was the direct cause of the following exception:

  File "/opt/ml/code/train.py", line 901, in <module>
    main()
  File "/opt/ml/code/train.py", line 897, in main
    run_experiment()
  File "/opt/ml/code/train.py", line 363, in __call__
    run_rsmtool_experiment(train_dev_rsmtool_config, train_dev_experiment_dir_path)
  File "/env/lib/python3.11/site-packages/rsmtool/rsmtool.py", line 336, in run_experiment
    ) = analyzer.run_prediction_analyses(new_pred_data_container, pred_config, wandb_run)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/env/lib/python3.11/site-packages/rsmtool/analyzer.py", line 1826, in run_prediction_analyses
    wandb.log_confusion_matrix(
  File "/env/lib/python3.11/site-packages/rsmtool/utils/wandb.py", line 151, in log_confusion_matrix
    f"{section}/{name}": wandb.plot.confusion_matrix(
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/env/lib/python3.11/site-packages/wandb/plot/confusion_matrix.py", line 78, in confusion_matrix
    counts[class_mapping[y_true[i]], class_mapping[preds[i]]] += 1
                         ~~~~~~^^^
  File "/env/lib/python3.11/site-packages/pandas/core/series.py", line 1040, in __getitem__
    return self._get_value(key)
           ^^^^^^^^^^^^^^^^^^^^
  File "/env/lib/python3.11/site-packages/pandas/core/series.py", line 1156, in _get_value
    loc = self.index.get_loc(label)
          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/env/lib/python3.11/site-packages/pandas/core/indexes/base.py", line 3797, in get_loc
    raise KeyError(key) from err
KeyError: 3
```

Let's talk more about this. I can probably provide you with a minimal case to reproduce the issue.

mulhod: I should say that so far, otherwise, everything is working as expected. I'm trying to wade through all of the stuff that gets logged to wandb to check things out.

mulhod: Ok, I think I know what is going on. Compare:

```
In [33]: human_scores.keys()
Out[33]: RangeIndex(start=0, stop=7317, step=1)

In [34]: human1_scores.keys()
Out[34]:
Index([   0,    1,    2,    4,    7,   15,   16,   18,   19,   21,
       ...
       7299, 7301, 7302, 7304, 7305, 7306, 7309, 7310, 7311, 7316],
      dtype='int64', length=3309)
```

For the first `Series`, `human_scores`, which the function call does not fail for without the `.values`, the index is just a range index. Any index value from 0 to 7,317 will work (that's how many items are in the series). But for the second `Series, `human1_scores` (and the same is true for `human2_scores`), which the function call _does_ fail for, the index of the series is basically a list of numbers that is a subset of `range(0, 7317)`. That is, some of the entries will not be there, notably `3` in my case. The function thinks the thing being passed in is a list-like thing that can be indexed in the normal way list-like things can be indexed. This works fine for the first series since that's true, but for the second type of series, it does not work. Simply adding `.values` converts the series to a `numpy` array, which can be indexed just like a list. I think, therefore, that all series should be converted to `numpy` arrays when calling the function. You could do that in the place I suggested above (but also a few lines up), or you could just do that in the `wandb.log_confusion_matrix` function:

```python
def log_confusion_matrix(wandb_run, human_scores, system_scores, name, section):
    """
    Log a confusion matrix to W&B if logging to W&B is enabled.
    The confusion matrix is added as a custom chart.
    Parameters
    ----------
    wandb_run : wandb.Run
        The wandb run object, or None, if logging to W&B is disabled
    human_scores : Sequence
        The human scores for the responses in the data
    system_scores : Sequence
        The predicted scores for the responses in the data
    name : str
        The chart title
    section : str
        The section in which the confusion matrix will we logged
    """
    if wandb_run:
        wandb_run.log(
            {
                name: wandb.plot.confusion_matrix(
                f"{section}/{name}": wandb.plot.confusion_matrix(
                    probs=None,
                    y_true=human_scores.values,
                    preds=system_scores.values,
                    title=name,
                )
            }
        )
```

I would probably suggest that the documentation be updated, however, to indicate that a `pd.Series` is expected.

mulhod: By the way, I subsequently removed my changes to `analyzer.py` and put in the changes to the `log_confusion_matrix` function that I suggested last (at the very bottom of my long comment) and I can confirm that that change fixed the bug.

mulhod: One thing I have noticed so far in my wading: The rows vs. columns seem to be inverted when comparing the WandB-logged confusion matrices with those in the reports. I was expecting the tables to look the same, but in fact they are inverted. We may want to invert the order of the arguments to the WandB-logged matrices (which are not labelled in a very nice way, anyway -- it's hard to tell which is human1 and which human2 or which is system and which is human). It appears to me that, at least if we invert them, they will be consistent.

tamarl08: I had similar issues with this method and also provided some details to [W&B](https://github.com/wandb/wandb/issues/5711). This is the reason that we don't have nice labels at the moment. 
I'll confirm that using `.values` work on our test data too.

inverting the order of arguments will work but will be a bit weird in the human-system case since then we do have predictions vs ground truth. 

We can also create the conf matrix plot on our own, at least until the bugs in that method are fixed.

tamarl08: Looks like maybe our matrix is inverted? the report says `(rows=system, columns=human)` which is like the wandb one. and we [transpose](https://github.com/EducationalTestingService/rsmtool/blob/main/rsmtool/analyzer.py#L1799) it for some reasone when creating the dataframe.

mulhod: Maybe we could invert the arguments to the function only for human-human. In that case, a change would be made in `analyzer.py`.

mulhod: > Looks like maybe our matrix is inverted? the report says `(rows=system, columns=human)` which is like the wandb one. and we [transpose](https://github.com/EducationalTestingService/rsmtool/blob/main/rsmtool/analyzer.py#L1799) it for some reasone when creating the dataframe.

Yeah, not sure anymore!

tamarl08: I counted in the raw data and indeed it is rows=human, columns=system.

tamarl08: yes.

desilinguist: Wow, I'm surprised no one's ever noticed. Actually, not that surprised since most people only look at the diagonals.

desilinguist (changes requested):
Actually, now that I think about it, we should mention the section stuff in documentation somewhere right?

desilinguist (approved):
LGTM!

mulhod (approved):

PR #662, opened 2023-10-04
Bug fixes in xval and report

tamarl08:
- fix the concatenation of prediction files from xval folds, which used the `keys` param incorrectly and unnecessarily. This caused a bug only with more than 9 folds.
- added a 10-fold scenario to the test of that method
- fixed the description of human-system confusion matrix to match the matrix axes values.
- changed the title of human-human confusion matrix to be different than the human-system one, so that the links at the top works correctly.

### DIFF
--- a/rsmtool/notebooks/consistency.ipynb
+++ b/rsmtool/notebooks/consistency.ipynb
@@ -89,7 +89,7 @@
     "                            formatters=formatter_dict, escape=False)))\n",
     "\n",
     "    if exists(confmat_h1h2_file):\n",
-    "        markdown_strs = ['### Confusion matrix']\n",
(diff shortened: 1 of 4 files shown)


rsmtool/notebooks/consistency.ipynb
```diff
@@ -89,7 +89,7 @@
     "                            formatters=formatter_dict, escape=False)))\n",
     "\n",
     "    if exists(confmat_h1h2_file):\n",
-    "        markdown_strs = ['### Confusion matrix']\n",
+    "        markdown_strs = ['### Human-human confusion matrix']\n",
```
mulhod, line 92: Oh, nice catch.

mulhod (approved):

desilinguist (approved):

PR #669, opened 2024-01-09
Refactor CLI tests to modernize codecov

### LINKED ISSUE #638: Refactor CLI tests to call `main()` functions instead of calling the entry points
CodeCov wants people to switch to their [binary uploader](https://docs.codecov.com/docs/codecov-uploader) instead of using the `codecov` package. However, in my tests (as part of #636), I found that when I switch to the new uploader, we were losing coverage – mainly because the [coverage subprocess measurement](https://coverage.readthedocs.io/en/7.2.7/subprocess.html) was not working as intended. 

Since we can't stay on the codecov python package forever, we will need to refactor our CLI tests so that instead of calling the entry point directly via `subprocess()`, we call the `main` function as we do in SKLL. 

There are two tasks in this issue: 
- [ ] Refactor CLI tests
- [ ] Switch to the binary CodeCov uploader

desilinguist:
- All tools main functions now take arguments. This is needed to refactor the CLI tests so that they call the main functions directly instead of using subprocess calls.
- Refactor the CLI tests since we want to be able to use the command-line `codecov` tool which does not measure subprocess coverage.
- Switch to `codecov` binary in the Gitlab CI plan and install `curl` so we can download that binary in the first place.
- Ignore cleanup errors in `test_cli.py`.

Notes:
1. A side-benefit of this change is that `test_cli.py` now runs 3x faster since we aren't using subprocesses.
2. The coverage change is expected since we are changing the total number of lines in the code.

Closes #638.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,11 +1,19 @@
+import logging
 import os
-import shlex
-import subprocess
-import sys
 import unittest
+from contextlib import redirect_stderr, redirect_stdout
+from io import StringIO
 from pathlib import Path
 from tempfile import NamedTemporaryFile, TemporaryDirectory
+from unittest.mock import patch
 
+from rsmtool.rsmcompare import main as rsmcompare_main
+from rsmtool.rsmeval import main as rsmeval_main
+from rsmtool.rsmexplain import main as rsmexplain_main
+from rsmtool.rsmpredict import main as rsmpredict_main
+from rsmtool.rsmsummarize import main as rsmsummarize_main
+from rsmtool.rsmtool import main as rsmtool_main
+from rsmtool.rsmxval import main as rsmxval_main
 from rsmtool.test_utils import (
     check_file_output,
     check_generated_output,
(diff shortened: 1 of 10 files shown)


tests/test_cli.py
```diff
@@ -54,30 +76,25 @@ class TestToolCLI(unittest.TestCase):
...
-            cmd = f"{binpath}/{context}"
-        else:
-            cmd = f"{context}"
+        # call the main function for the given context
+        # with no arguments and check that it exits
+        # with exit code 0 and prints out the usage
+        main_function = MAIN_FUNCTIONS[context]
+        with redirect_stdout(StringIO()) as captured_stdout:
+            with self.assertRaises(SystemExit) as excm:
+                main_function([])
```
tamarl08, line 91: shouldn't we test it with no params, like that: `main_function()`, then argue is None and it will cover the lines above with the warnings?

desilinguist: Yes,  I tried that but the problem is if `argv` is `None`, it relies on `sys.argv` which is then going to be the `nose2` arguments 😉.

desilinguist: This test case does test what'd happen if we just run `rsmtool` without any arguments since the `rsmtool` entry point points to `rsmtool.rsmtool.main`.

tamarl08: but will it be `None`, or `[]` if we run it with no arguments? If it's an empty list, when will it be `None`?

desilinguist: Yes, it will be `None` but since `sys.argv[1:]` will be an empty list, it's the same as passing `[]`.  See first two lines in `rsmtool.main()`.

tamarl08: oh, got it now!

tamarl08 (approved):

PR #677, opened 2024-01-29
Add type hints (part 1)

desilinguist:
- Update pre-commit checks and add mypy
- Add type hints to `rsmtool.py`.
- Add type hints to `rsmeval.py`
- Add type hints to `rsmpredict.py`
- Add type hints to `rsmcompare.py`
- Add type hints to `rsmsummarize.py`
- Add type hints to `rsmxval.py`
- Add type hints to `rsmexplain.py`
- Add type hints to `convert_feature_json.py`
- Add `wandb` as mypy dependency


mulhod (approved):

rsmtool/convert_feature_json.py
```diff
@@ -15,7 +16,7 @@
 import pandas as pd
 
 
-def convert_feature_json_file(json_file, output_file, delete=False):
+def convert_feature_json_file(json_file: str, output_file: str, delete=False) -> None:
```
tamarl08, line 19:
```suggestion
def convert_feature_json_file(json_file: str, output_file: str, delete: bool=False) -> None:
```

rsmtool/rsmeval.py
```diff
@@ -28,8 +32,12 @@
 
 
 def run_evaluation(
-    config_file_or_obj_or_dict, output_dir, overwrite_output=False, logger=None, wandb_run=None
-):
+    config_file_or_obj_or_dict: Union[str, Configuration, Dict[str, Any], Path],
+    output_dir: str,
+    overwrite_output=False,
```
tamarl08, line 37:
```suggestion
    overwrite_output: bool =False,
```

rsmtool/rsmexplain.py
```diff
@@ -133,11 +142,11 @@ def mask(learner, featureset, feature_range=None):
 
 
 def generate_explanation(
-    config_file_or_obj_or_dict,
-    output_dir,
+    config_file_or_obj_or_dict: Union[str, Configuration, Dict[str, Any], Path],
+    output_dir: str,
     overwrite_output=False,
```
tamarl08, line 147:
```suggestion
    overwrite_output: bool =False,
```

rsmtool/convert_feature_json.py
desilinguist: Yeah, @mulhod asked me the same thing :) I was avoiding adding hints to things that were explicitly and obviously boolean but since both of you suggested this, I think I should just do it. Will fix.

tamarl08: I realized at some point that it's not a typo.. it is obvious but it's not the only obvious thing :)
how did mypy not complain?

desilinguist: It doesn't complain because it can infer the type successfully without the hint :)

tamarl08 (approved):

PR #688, opened 2024-09-20
648/add excluded features

### LINKED ISSUE #648: Add list of excluded features to the report
Excluded feature columns (e.g. for non numeric values) are only printed to STDOUT- would be helpful to add a table of these column to the data description section of the report.

damien2012eng:
closes #648 

The PR resolves two issues:
- Add a condition to check if `df_filtered` is empty after filtering out from previous iteration.
- Include the zero SD features message in the report. 

How to test:
1. Check to this branch. 
2. cd to the `examples/rsmtool`. Add/modify a/more new columns with all values as string to the `train.csv`. 
3. Run rsmtool command. 
4. Check the message on the command, ensure only the new added/modified columns are printed. 
5. Revert the changes in the `train.csv`
6. Add a/more new columns with all values being 1. 
7. Run rsmtool commmand. 
8. Check the final report to ensure that the zero excluded feature messages are added in the top of the report.

### DIFF
--- a/rsmtool/notebooks/data_description.ipynb
+++ b/rsmtool/notebooks/data_description.ipynb
@@ -15,10 +15,7 @@
    "source": [
     "# print the excluded features in the report if exists.\n",
     "if excluded_features:\n",
-    "    if len(excluded_features) == 1:\n",
-    "        display(Markdown(f\"The following feature was excluded because its standard devision in the training set is equal to 0: **{excluded_features[0]}**.\"))\n",
-    "    else:\n",
-    "        display(Markdown(f\"The following features were excluded because their standard devision in the training set is equal to 0: **{\", \".join(excluded_features)}**.\"))"
+    "    display(Markdown(f\"The following feature(s) were excluded because their standard devision in the training set is equal to 0: **{\", \".join(excluded_features)}**.\"))"
    ]
   },
   {
--- a/rsmtool/notebooks/data_description.ipynb
+++ b/rsmtool/notebooks/data_description.ipynb
@@ -13,6 +13,20 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "# print the excluded features in the report if exists.\n",
+    "if excluded_features:\n",
+    "    if len(excluded_features) == 1:\n",
+    "        display(Markdown(f\"The following feature was excluded because its standard devision in the training set is equal to 0: **{excluded_features[0]}**.\"))\n",
+    "    else:\n",
+    "        display(Markdown(f\"The following features were excluded because their standard devision in the training set is equal to 0: **{\", \".join(excluded_features)}**.\"))"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
     "try:\n",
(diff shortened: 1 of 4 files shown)


desilinguist:
@damien2012eng @tamarl08 so just to confirm:

1. The first test with an all-string feature should result is one excluded feature message and then an error saying that there are no responses remaining.
2. The second tests with an all-1 feature should result in a warning with the feature excluded message, a successful RSMTool run, and then a message at the top of the report saying that the feature was excluded?

rsmtool/notebooks/header.ipynb
```diff
@@ -233,7 +234,11 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "Markdown('''This report presents the analysis for **{}**: {}'''.format(experiment_id, description))"
+    "display(Markdown('''This report presents the analysis for **{}**: {}'''.format(experiment_id, description)))\n",
+    "# print the excluded features in the report if exists.\n",
+    "if excluded_features:\n",
+    "    for feature in excluded_features:\n",
+    "        display(Markdown(f\"Feature **{feature}** was excluded from the model because its standard deviation in the training set is equal to 0.\"))"
```
desilinguist, line 241: Couple of thoughts on this?

1. Is this the best place for this or should this go in one of the sections? I know that will be extra work so may be the is fine?
2. Instead of repeating the same sentence for each feature, can we just have something like: `The following features were excluded because their standard devision in the training set is equal to 0: F1, F2, F3`?

tamarl08:
@damien2012eng  I tested and it works as expected!
I think the list of features should go in the `data description` section. if it is too complicated, let's try and make it look like a separate section in the header (e.g. add a paragraph title or a "Note" box)

damien2012eng:
@desilinguist @tamarl08  I moved the message to the `data description` section. Second round?

rsmtool/notebooks/data_description.ipynb
```diff
@@ -7,6 +7,20 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# print the excluded features in the report if exists.\n",
+    "if excluded_features:\n",
+    "    if len(excluded_features) == 1:\n",
+    "        display(Markdown(f\"The following feature was excluded because its standard devision in the training set is equal to 0: **{excluded_features[0]}**.\"))\n",
+    "    else:\n",
+    "        display(Markdown(f\"The following features were excluded because their standard devision in the training set is equal to 0: **{\", \".join(excluded_features)}**.\"))"
+   ]
```
desilinguist, line 22: I don't think you need an if statement.  Just say: `The following feature(s) were excluded because their standard devision in the training set is equal to 0:`. That should work in either case. 

Also, I am getting an error when I run this – something about a missing `}`.

desilinguist (changes requested):

rsmtool/notebooks/data_description.ipynb
damien2012eng: hmm. Interesting. May I ask how did you run it? I didn't see a missing }.

desilinguist (approved):
LGTM!

tamarl08 (approved):

PR #691, opened 2025-01-02
Example to export model to ONNX

### LINKED ISSUE #689: Slim installation option or model inference export option
rsmtools has many large dependencies. Many of them are not needed at inference time.

It would be nice to either:

1. Use the pip compatible install options like `pip install "rsmtools[full]"` to install all dependencies only when needed
2. Add a method to export an already trained model so that it can be run without requiring rsmtools (for example as a sklearn pipeline). Ideally it would allow specifying all the options that `fast_predict` supports.

twoertwein:
Closes #689

I added an example to demonstrate that it is possible to convert an RSMTool model to ONNX.

I just implemented the transformations to make the conversion work for the model I was dealing with :) The example will error when it encounters skll/rsmtool features that it does not (yet) support/when the conversion fails.

I was dealing with a simple regression model. `fast_predict` takes 0.1s while the converted ONNX+numpy model takes only 0.00014s

### DIFF
--- a/doc/advanced_usage.rst
+++ b/doc/advanced_usage.rst
@@ -16,3 +16,5 @@ In addition to providing the ``rsmtool`` utility training and evaluating regress
 .. include:: usage_rsmxval.rst.inc
 
 .. include:: usage_rsmexplain.rst.inc
+
+.. include:: usage_onnx_deployment.rst.inc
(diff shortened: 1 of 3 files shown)


desilinguist:
This is great @twoertwein! We will also need to add details about this to the documentation explaining the use case and the dependencies needed (e.g, `skl2onnx` etc.).

twoertwein:
> We will also need to add details about this to the documentation explaining the use case and the dependencies needed (e.g, `skl2onnx` etc.).

In the form of a quick tutorial like this one https://github.com/EducationalTestingService/rsmtool/blob/main/doc/tutorial_rsmcompare.rst.inc or should this be documented differently?

desilinguist:
A tutorial would be fantastic!

